PR #5, opened 2016-05-12
Added image proxy extension

Frenzie:


Alkarex:
Thanks :+1:  This looks interesting.
It would be good to have the possibility of having the FreshRSS server do the proxying itself, without necessarily using a third-party service. If you can, please consider adding an option, or maybe simply a readme explaining the choice of the proxy service and possibility of self-hosting.

Frenzie:
Good point. How about something like this?

Frenzie:
Would you like me to change anything about the README.md?

Alkarex:
Sorry, I had forgotten about this pull request...
Thanks!

PR #7, opened 2016-09-19
Add CSRF token, change ImageProxy type from user to system, and add CustomJS extension

Frenzie:
This is for https://github.com/FreshRSS/FreshRSS/issues/1253

CustomJS is CustomCSS' natural counterpart.

### DIFF
--- a/xExtension-CustomJS/metadata.json
+++ b/xExtension-CustomJS/metadata.json
@@ -1,0 +1,8 @@
+{
+	"name": "Custom JS",
+	"author": "₣rans de Jonge",
+	"description": "Apply custom JS.",
+	"version": 0.1,
+	"entrypoint": "CustomJS",
+	"type": "user"
+}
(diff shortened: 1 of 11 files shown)


xExtension-CustomJS/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "Custom JS",
+	"author": "Marien Fressinaud",
```
Alkarex: You should have your own name there @Frenzie :-)

Alkarex:
Looks very good, thanks :+1: 
Did you check that it works with the additional CSRF token?

xExtension-CustomJS/metadata.json
Frenzie: I didn't really write it but if you say so. =P

Frenzie:
Yeah, CustomCSS, CustomJS and ImageProxy are all working fine. =)

PR #12, opened 2016-12-04
Added third-party extension list

### LINKED ISSUE #11: Place for listing FreshRSS extensions
Hi there, 
while searching for FreshRSS extensions I stumbled upon this repository, but couldn't find any other resource.
I would suggest to have a place where users can link their own extensions.
This is not only a shameless try to drop a link to my [freshrss youtube extension](https://github.com/kevinpapst/freshrss-youtube) ;-) but I believe that the FreshRSS community would benefit from it.

The easiest way for the start could be to include a new section in this repositories README with links to "third party extensions". I could sent in a PR if you like the idea.

Any other suggestions?

kevinpapst:
Fixed #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,10 @@
 This repository contains all the official FreshRSS extensions.
 
 To install an extension, download [the extension archive](https://github.com/FreshRSS/Extensions/archive/master.zip) first and extract it on your PC. Then, upload the specific extension(s) you want on your server. Extensions must be in the `./extensions` directory of your FreshRSS installation.
+
+## Third-party extensions
+
+There are some FreshRSS extensions out there, which were developed by community members:
+
+- [Youtube](https://github.com/kevinpapst/freshrss-youtube) - will display videos from YouTube feeds inline
+


Alkarex:
Would you like to keep your extension in your own repository? You are otherwise welcome to make a pull request to https://github.com/FreshRSS/Extensions

kevinpapst:
To be honest: If this would be my project, I would create one repository for each extension, so its easier to version them. Then I would create a page (or README section) with links and a short description to each of them. Its easier to find maintainers for each plugin this way - and it might even be more motivating for other Devs to join the community if they see that you can create extensions easily.

Alkarex:
We currently do not have one repository per extension, but I think we have a good subset of what you ask: a readme for each extension (e.g. https://github.com/FreshRSS/Extensions/tree/master/xExtension-ImageProxy ), and their own version. What we do not have is individual commit rights for each extension. But I would have no problem giving you write access to https://github.com/FreshRSS/Extensions if you host an extension there.

PR #14, opened 2017-03-17
Update README.md

oyox:
Added the oYoX extension collection


Alkarex:
Thanks for this contribution 👍  I will have a look.
We could maybe consider promoting some of those features to FreshRSS core.

PR #22, opened 2017-12-03
Extension list JSON file

kevinpapst:
This file contains all currently existing extensions, required for this proposal: 
https://github.com/FreshRSS/FreshRSS/pull/1708

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1,0 +1,61 @@
+{
+  "version": 0.1,
+  "extensions": [
+    {
+      "name": "CustomCSS",
+      "description": "A FreshRSS extension which give ability to create user-specific CSS rules to apply in addition of the actual theme",
+      "version": 0.1,
+      "author": "Marien Fressinaud",
+      "url": "https://github.com/FreshRSS/Extensions",
+      "type": "gh-subdirectory"
+    },
+    {
+      "name": "CustomJS",
+      "description": "A FreshRSS extension which give ability to create user-specific Javascript rules to apply in addition of the actual theme",
+      "version": 0.1,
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -0,0 +1,117 @@
...
+      "version": 0.1,
+      "author": "Marien Fressinaud",
+      "url": "https://github.com/FreshRSS/Extensions",
+      "type": "gh-subdirectory"
+    },
+    {
+      "name": "CustomJS",
+      "description": "A FreshRSS extension which give ability to create user-specific Javascript rules to apply in addition of the actual theme",
+      "version": 0.1,
+      "author": "₣rans de Jonge",
```
Frenzie: Is it just my phone or did you use some odd glyph instead of F?

kevinpapst: I double checked that yesterday as well as it looked unfamiliar to me as well ... but it comes from https://github.com/FreshRSS/Extensions/blob/master/xExtension-CustomJS/metadata.json

Frenzie: I… typod my own name with the French franc? O_o That's a new one.

I guess I must've accidentally used some Compose key sequence.

kevinpapst: I changed your name in this file, but the metada.json is still open

Alkarex:
We should remove the TT-RSS extension. It is not functional as it is. I have started an upgrade https://github.com/FreshRSS/FreshRSS/pull/1526 but it requires some database changes, which I plan to do for FreshRSS 1.10.

kevinpapst:
@Alkarex removed

PR #29, opened 2018-04-06
upgraded youtube to 0.6

kevinpapst:


Frenzie:
YouTube channel feeds? YouTube offers feeds without rss-bridge somewhere?

kevinpapst:
Yes, it does, e.g. 
https://www.youtube.com/feeds/videos.xml?channel_id=UCuP2vJ6kRutQBfRmdcI92mA

The link for a user is slightly different, but also works .

Frenzie:
Huh, good to know. I wonder why it keeps them so well hidden, because I can't find those.

Doesn't look like rss-bridge has any notion of these feeds either: https://github.com/RSS-Bridge/rss-bridge/blob/master/bridges/YoutubeBridge.php

Edit: oh, I'll be damned!

screenshot_2018-04-06_20-34-45

That's some roundabout way of getting at. For a decade I've thought they removed the feature but I guess it might've been hiding out there all this time.

What are the other link variations? Obvious stuff like `username=string`?

kevinpapst:
There are actually a couple of feeds. One no your subscription page (if you have a youtube account) right at the bottom there is an OPML file.
And besides the one from my previous posting there are also these:
https://www.youtube.com/feeds/videos.xml?user=xxx
https://www.youtube.com/feeds/videos.xml?playlist_id=xxxx

Frenzie:
> One on your subscription page (if you have a youtube account) right at the bottom there is an OPML file download link.

Which subscription page is that? Are you talking about the one I just posted a screenshot of? Because the regular one is endless scrolling. No bottoms there. :-P

kevinpapst:
I am not using Youtube native subscriptions, as I only use FreshRSS for that, but I meant this page:
https://www.youtube.com/subscription_manager

PR #33, opened 2018-07-01
New extension : add a Reading Time estimation

lapineige:
See: https://github.com/FreshRSS/FreshRSS/issues/1260


kevinpapst:
Hi @lapineige,
please add it to https://github.com/FreshRSS/Extensions/blob/master/extensions.json as well, then it will be displayed in the admin area as available extension.

lapineige:
Ah, good point !
Thanks, I didn't know that was possible.

I'll do it :)

> Important: the name field must be the same than entrypoint in your metadata.json!
(why ?)

Ok, I'll correct that.

kevinpapst:
Sorry for the misunderstanding, I meant the `name` in extensions.json must be the same than `entrypoint` in metadata.json.
But the `name` in metadata.json is up to you. So you can change it back.

lapineige:
Well, it doesn't really matter 😄. I leave it like this.

xExtension-ReadingTime/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 Lapineige
```
kevinpapst: ;-)

xExtension-ReadingTime/README.md
```diff
@@ -0,0 +1,19 @@
+Extension pour FreshRSS (https://github.com/FreshRSS)
+
+**v1.0**
+
+Ajoute une estimation du temps de lecture à côté de chaque article.
+Fonctionne sur les affichages de bureau et mobile.
+
+Source: https://framagit.org/Lapineige/FreshRSS_Extension-ReadingTime
```
kevinpapst: So you are developing it at framagit.org and copy changes to this repository here?

kevinpapst:
@Alkarex do you prefer to have new extensions hosted in this "official FreshRSS extensions" repository or in their own repository or it doesn't matter?

xExtension-ReadingTime/LICENSE
lapineige: ? (I don't understand ^^)

xExtension-ReadingTime/README.md
lapineige: Yes. I prefer to keep my files outside of github.
I'll copy the changes (I don't plan to update it that often ^^).

xExtension-ReadingTime/LICENSE
lapineige: Well, it seems that I started this a while ago ^^
(end of 2016)

Alkarex:
Sorry for the delay (nice weather and all that...). I do not have clear preferences, and it depends on the cases. It could be a kind of progression to first start in their own repositories, and when getting more used / important, to move to this repository, especially when there is more than one developer contributing.

xExtension-ReadingTime/LICENSE
Alkarex: I often write something like `2016 - 2018` in such a situation

xExtension-ReadingTime/README.md
Alkarex: I believe git hooks could be used there to automatically keep two repositories in perfect sync.

Alkarex:
@lapineige I have made a few corrections, in particular in the event registration (the script would fail if it was slower than DOMContentLoaded, or faster than jQuery init). Please double-check before we merge.

Here is how it looks on my instance:

image

xExtension-ReadingTime/README.md
lapineige: Does it work from gitlab to github ?

Alkarex:
When the screen is too small, then the feed titles are not shown, so your reading time disappears as well.

lapineige:
It should be shown in the title of the article now (you need to reload the page).

lapineige:
> Please double-check before we merge.

It works as expected :)

xExtension-ReadingTime/README.md
Alkarex: I have not tried, but I believe so, yes

PR #34, opened 2018-07-22
Fix parse error on extensions.json

Leepic:
Related issue [#1965 on FreshRSS/FreshRSS](https://github.com/FreshRSS/FreshRSS/issues/1965)


kevinpapst:
Thanks for the fix @Leepic

PR #36, opened 2018-09-01
Add a "Force proxying, even for https" option

minami-o:
Hi there,
I added a checkbox to ImageProxy in order to enforce the use of the proxy, even for images coming from https sources.
The intention is to make the FreshRSS host the only point of contact for the client, preventing direct connection to the RSS sources which could be a privacy concern, or could be blocked by browser addons (for instance µMatrix, when blocking external images except whitelisted hosts).
Not sure anyone but me is interested, but well… :)
Please let me know if anything should be amended to comply with the extensions' rules.
Cheers!


minami-o:
Hi,
Not at all, I configured vim for python so my tabs were 4 spaces, which was inconsistent with the current code.
When I realized this, I used "%s/    /\t/" and didn't see the impact on this line :)
Cheers

1 septembre 2018 06:19 "Frans de Jonge" <notifications@github.com> a écrit:

Frenzie:
Alright, thanks for the addition! If you could remove the final whitespace changes and also edit this:

https://github.com/FreshRSS/Extensions/blob/b379403571823bda095c163d5d4589692cb9fa26/extensions.json#L21-L26

then I'll merge it. :-)

xExtension-ImageProxy/extension.php
```diff
@@ -3,7 +3,7 @@
 class ImageProxyExtension extends Minz_Extension {
 	public function init() {
 		$this->registerHook('entry_before_display',
-		                    array('ImageProxyExtension', 'setImageProxyHook'));
+                            array('ImageProxyExtension', 'setImageProxyHook'));
```
Frenzie: It should either be one tab or however many spaces it was. ;-)

Frenzie: PS Note that's two tabs followed by alignment spaces.

minami-o: lol noticed that, trying to copy-paste this correctly from the initial code (that's what I thought I'd done, but looks I missed stg :D)

minami-o:
Thank you for the original piece!

2 septembre 2018 12:22 "Frans de Jonge" <notifications@github.com> a écrit:

PR #47, opened 2019-04-06
KeepFolderState

Kaan88:
Keep folder state is supposed the keep the folders open like this, even if you refresh the page or go to another folder.

image
image

Now even though the arrow shows that the folder is open it's actually closed.

image
image


Alkarex:
Let's see if we put it in core https://github.com/FreshRSS/FreshRSS/issues/1457 or leave it as an extension

PR #55, opened 2019-08-13
Title wrap extension

jkinable:
Implements a light-weight extension which applies a line-wrap to long titles as opposed to truncating the title. The css used to implement this behavior was originally proposed by @Frenzie in this [issue](https://github.com/FreshRSS/FreshRSS/issues/2344).

Obviously, the same can be achieved by enabling the xExtension-CustomCSS, and appending the aforementioned css code via the plugin management interface. However, not only is this not as obvious to new users, it is also harder to enforce this as the default behavior for newly created users. This new plugin can simply be added to the 'extensions_enabled' configuration option in config-user.default.php

### DIFF
--- a/xExtension-TitleWrap/static/.gitignore
+++ b/xExtension-TitleWrap/static/.gitignore
@@ -1,0 +1,1 @@
+style.*.css
--- a/xExtension-TitleWrap/static/.gitignore
+++ b/xExtension-TitleWrap/static/.gitignore
@@ -1,1 +1,0 @@
-style.*.css
--- a/xExtension-TitleWrap/static/title_wrap.css
+++ b/xExtension-TitleWrap/static/title_wrap.css
@@ -1,23 +1,21 @@
 .horizontal-list {
- display: flex;
- /* uncomment to allow title to expand onto its own line */
- /*flex-wrap: wrap;*/
+    display: flex;
 }
 .horizontal-list.bottom {
- display: table;
+    display: table;
 }
 .flux .item {
- flex-shrink: 0;
- line-height: normal;
- padding: .3em 0;
+    flex-shrink: 0;
+    line-height: normal;
+    padding: .3em 0;
 }
 .flux .item > a {
- white-space: normal;
+    white-space: normal;
 }
 .flux:not(.current):hover .item.title {
- position: relative;
- max-width: inherit;
+    position: relative;
+    max-width: inherit;
 }
 .flux_header .title {
- flex: auto;
+    flex: auto;
 }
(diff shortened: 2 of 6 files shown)


jkinable:
Note: I've no idea which license to use, so I just copy-pasted the license from one of the other extensions. It might make more sense to release all extensions under the same license, i.e. use a single license as opposed to an individual license per extension.

Frenzie:
On the one hand it seems slightly odd that all (?) the extensions have the same license as the top-level repo, on the other hand extensions might be distributed by themselves and then lose their license. @marienfressinaud was that the thought behind it?

As an extension this seems basically fine to me as is.

marienfressinaud:
I didn't realize that there is a LICENSE file in the root directory, and I don't remember why it was done this way.

My preference goes to having each extension with its own license for the same reasons as yours @Frenzie (extensions might be distributed separately). Also, authors are credited and they should have the right to choose their license (while it's "free"). I'm just not sure how the AGPL license (which applies to the core) acts upon the extensions… but as a maintainer, I would be ok to integrate a "MIT" extension for example.

Anyway, if the AGPL is fine for you @jkinable, let's go with it :)

I'll take a look to your PR a bit later.

jkinable:
@marienfressinaud I'm fine with the license (especially since I didn't read it...). I'd recommend to include a license field in the `metadata.json` file as well.

xExtension-TitleWrap/static/.gitignore
```diff
@@ -0,0 +1 @@
+style.*.css
```
marienfressinaud: you don't need this file

xExtension-TitleWrap/static/title_wrap.css
```diff
@@ -0,0 +1,23 @@
+.horizontal-list {
+ display: flex;
+ /* uncomment to allow title to expand onto its own line */
+ /*flex-wrap: wrap;*/
```
marienfressinaud: It's kind of useless to have this information here and it will be erased during an update. Since it can be set with the CustomCSS extension, you can give this CSS rule as an example in the README.

xExtension-TitleWrap/static/title_wrap.css
```diff
@@ -0,0 +1,23 @@
...
+.flux .item > a {
+ white-space: normal;
+}
+.flux:not(.current):hover .item.title {
+ position: relative;
+ max-width: inherit;
+}
+.flux_header .title {
+ flex: auto;
+}
```
marienfressinaud: Just a note about indentation: I recommend to use 1 tabulation (i.e. FreshRSS coding style) or 4 spaces (my current preference). 1 space is a bit odd ^^

PR #59, opened 2019-10-03
Update CustomCSS metadata.json to version 0.2

BrookeDot:
When installed the extension will show there is an update available because this file says version `0.1` where the main [`extensions.json`](https://github.com/FreshRSS/Extensions/blob/master/extensions.json) shows version `0.2` is the latest which is inline with the extension README


Alkarex:
Thanks 👍

PR #60, opened 2019-10-20
Fix non working extension with customised article top bar 

### LINKED ISSUE #58: [xExtension-ReadingTime] doesn't work unless "Date of publication" is checked
The time doesn't display at all (to the left of the article title) when "Date of publication" is not checked in the Display settings for the Article icons for the top line.

lapineige:
Fix #58

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -36,7 +36,7 @@
     {
       "name": "YouTube",
       "description": "Display videos from YouTube and PeerTube feeds inline",
-      "version": 0.9,
+      "version": 0.10,
       "author": "Kevin Papst",
       "url": "https://github.com/kevinpapst/freshrss-youtube",
       "type": "gh-subdirectory"


xExtension-ReadingTime/static/readingtime.js
```diff
@@ -23,15 +23,15 @@
          reading_time.words_count = reading_time.flux_words_count(flux_list[i]); // count the words
          reading_time.reading_time = reading_time.calc_read_time(reading_time.words_count, 300); // change this number (in words) to your prefered reading speed
 
+
```
Frenzie: Stray newline?

xExtension-ReadingTime/static/readingtime.js
```diff
@@ -42,18 +42,20 @@
 
  flux_words_count: function flux_words_count(flux) {
 
-     reading_time.textContent = flux.textContent; // get textContent
+     reading_time.textContent = flux.childNodes[2].childNodes[1].textContent; // get textContent, from the article itself (not the header, not the bottom line). `childNodes[2].childNodes[1]` gives : `<div class="content no_limit">` element
```
Frenzie: Seems a bit fragile. Would it be easier and less likely to change to do something like flux.querySelector('.content.no_limit')?

xExtension-ReadingTime/static/readingtime.js
lapineige: I don't understand, sorry ^^

xExtension-ReadingTime/static/readingtime.js
lapineige: That's right.
I tried that, but I couldn't get a result with that command (I don't understand why).
I could also just text the full text as before, then subtract an arbitrary value, the accuracy isn't such a big deal…

xExtension-ReadingTime/static/readingtime.js
Frenzie: I mean there was one newline and now there are two for no apparent reason. :-)

xExtension-ReadingTime/static/readingtime.js
Frenzie: Either way is fine by me, just trying to think along. ;-)

xExtension-ReadingTime/static/readingtime.js
```diff
@@ -23,15 +23,15 @@
...
          reading_time.reading_time = reading_time.calc_read_time(reading_time.words_count, 300); // change this number (in words) to your prefered reading speed
 
+
          if (document.body.clientWidth <= 840) { // in mobile mode, the feed name is not visible (there is only the favicon)
              // add the reading time right before article's title
              // in that case, [Time] - [Title] format is used instead of a "|" (as it looks better and doesn't take much more space)
-             if ( document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.title a").textContent.substring(0,(reading_time.reading_time + 'm - ').length) != reading_time.reading_time + 'm - ' ) {
-                 document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.title a").textContent = reading_time.reading_time + 'm - ' + document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.title a").textContent;
+             if ( $("#" + reading_time.flux.id + " ul.horizontal-list li.item.title a")[0].textContent.substring(0,(reading_time.reading_time + 'm - ').length) != reading_time.reading_time + 'm - ' ) {
```
Alkarex: It would be a pity to reintroduce jQuery. What is the problem with `document.querySelector()` ?

xExtension-ReadingTime/static/readingtime.js
lapineige: Oh, yeah, probably a comment line that I did not removed completely ^^
(does it matters ? ^^)

xExtension-ReadingTime/static/readingtime.js
Alkarex: We have removed jQuery from FreshRSS, except for the statistics page, so if you really need it, you need to explicitly load it

xExtension-ReadingTime/static/readingtime.js
lapineige: I way thinking this way: that's a quick fix, still a bit better in accuracy than the "arbitrary value", it's a bit dirty but working (and I didn't really have the time to investigate further), if it breaks again (I'll see it, I'm using it) I'll change it to the arbitrary value (or have more time to properly do it).
I know, it's dirty ^^
I'm ok to change it if you want.

xExtension-ReadingTime/static/readingtime.js
lapineige: Oh, dammit… it seems I did a mistaken git pull from an outdated repo… I didn't see that, that line should not be like that…

Alkarex:
@lapineige I have not read the full thread: what DOM elements are you trying to get?

lapineige:
The article content, without the top and bottom bar around it.

Alkarex:
`.querySelector('.flux_content')` should do it then

lapineige:
I tried, it didn't work. I don't know why.

xExtension-ReadingTime/static/readingtime.js
lapineige: Fixed :)

Alkarex:
Can you should your code trying `.querySelector('.flux_content')` ?
That class is there especially for that

lapineige:
Well, I tried again, this time in the real script and not the console… it worked.

xExtension-ReadingTime/static/readingtime.js
```diff
@@ -42,18 +42,20 @@
...
+     // flux.childNodes[0].childNodes[2].textContent.split(' ').length - // excluding website length
+     // flux.childNodes[0].childNodes[4].textContent.split(' ').length - // excluding date length
+     // flux.childNodes[0].childNodes[6].textContent.split(' ').length;  // excluding title length
+     // Not 100% acurate as it doesn't check if there are shown at the bottom of the article
      return reading_time.textContent.split(' ').length - //raw number of words
-     flux.childNodes[0].childNodes[2].textContent.split(' ').length - // excluding website length
-     flux.childNodes[0].childNodes[4].textContent.split(' ').length - // excluding date length
-     flux.childNodes[0].childNodes[6].textContent.split(' ').length;  // excluding title length
-     //Not 100% acurate as it doesn't check if there are shown at the bottom of the article
+     15 ; // Temporary fix for excluding website/date/title text. It assumes than the title (+ author) label is rougly 15 words long.
```
Alkarex: What is that? :-)

lapineige: The article content includes article title, name of the author, sometimes the date…
It's not "real" content that will be read.
So it can increase a bit the estimated reading time (but it's not a big deal).
As I don't know how to exclude those elements, I've just decided to use an arbitrary value¹ (15 words) to correct that bias.

¹ based on the average from a selection of some articles.

Alkarex: Sorry, I did not see the `-` on the previous line

xExtension-ReadingTime/static/readingtime.js
```diff
@@ -42,18 +42,20 @@
 
  flux_words_count: function flux_words_count(flux) {
 
-     reading_time.textContent = flux.textContent; // get textContent
+     reading_time.textContent = flux.querySelector('.flux_content').textContent; // get textContent, from the article itself (not the header, not the bottom line). `childNodes[2].childNodes[1]` gives : `<div class="content no_limit">` element
```
marienfressinaud: You should use `'.flux_content .content'` to be sure to get only the content to read

xExtension-ReadingTime/static/readingtime.js
```diff
@@ -42,18 +42,20 @@
...
      // split the text to count the words correctly (source: http://www.mediacollege.com/internet/javascript/text/count-words.html)
      reading_time.textContent = reading_time.textContent.replace(/(^\s*)|(\s*$)/gi,"");//exclude  start and end white-space
      reading_time.textContent = reading_time.textContent.replace(/[ ]{2,}/gi," ");//2 or more space to 1
      reading_time.textContent = reading_time.textContent.replace(/\n /,"\n"); // exclude newline with a start spacing
 
+     /// This part ↓↓↓ is not working if the content of the title bar is changed by the user… need to do that in proper, more robust way
+     // flux.childNodes[0].childNodes[2].textContent.split(' ').length - // excluding website length
+     // flux.childNodes[0].childNodes[4].textContent.split(' ').length - // excluding date length
+     // flux.childNodes[0].childNodes[6].textContent.split(' ').length;  // excluding title length
+     // Not 100% acurate as it doesn't check if there are shown at the bottom of the article
```
marienfressinaud: No need to explain all of this, since the header is no longer returned by the querySelector

xExtension-ReadingTime/static/readingtime.js
marienfressinaud: and no need of this temporary fix, just return `reading_time.textContent.split(' ').length`

xExtension-ReadingTime/static/readingtime.js
marienfressinaud: It just makes the code more clear ;)

marienfressinaud:
Not sure what I did :thinking: I think that GitHub doesn't like when we work on the master branch of fork repositories. I'll reopen a PR with my own fork

marienfressinaud:
Should be merged now :)

lapineige:
Thanks a lot, sorry for not answering.

marienfressinaud:
No problem, I saw the :rocket: emoji so I considered it was good for you ^^

PR #63, opened 2019-12-07
Add "Share By Email" extension

marienfressinaud:
This is a very old request (november 2013!) to be able to share an
article directly from FreshRSS. I don't think it should be within the
core of FRSS, so I played a bit with the extension system. I keep it
very simple for now, just to try to see what was possible with the
current system. I just needed to add a new placeholder `(~ID~)` to
FreshRSS_Share.

It relies on the mailing system introduced for the "email validation"
feature.

Reference: https://github.com/FreshRSS/FreshRSS/pull/2707

Closes https://github.com/FreshRSS/FreshRSS/issues/300

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -152,6 +152,14 @@
       "author": "CNTools | Clemens Neubauer",
       "url": "https://github.com/cn-tools/cntools_FreshRssExtensions",
       "type": "gh-subdirectory"
+    },
+    {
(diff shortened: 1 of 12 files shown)


marienfressinaud:
Ping @aledeg ;)

aledeg:
I have to review that :)

PR #64, opened 2019-12-11
add infos about extensions from cn-tools

cn-tools:
It would be nice if you accept it.
It is promotion around my own extensions for FreshRSS only.

PR #66, opened 2019-12-19
add infos about extensions from cn-tools

cn-tools:
I hope i have changed the documentation from 'denglish' into 'english' correct!

Ref. #64 

Ref: FreshRSS/FreshRSS#2704


marienfressinaud:
I'm merging this PR, but just to let you know @cn-tools that we merged a PR yesterday in https://github.com/kevinpapst/freshrss-youtube/ extension. It does the same thing that your "YouTube Channel 2 RSSFeed" extension, but it's a bit more robust so you might be interested by the implementation ([here](https://github.com/kevinpapst/freshrss-youtube/blob/master/xExtension-YouTube/extension.php#L47-L60)).

PR #71, opened 2020-01-20
Add Quick Collapse extension

marienfressinaud:
This PR brings code from https://github.com/FreshRSS/FreshRSS/pull/1555 into an extension.

I had to be a bit creative because there's no hooks to add buttons to the "nav_menu". I decided to try to add it in JavaScript to avoid the immediate need of a hook. Unfortunately, it was harder to generate the icons URLs and title translations since we can't use the PHP helpers: I had to generate JS variables from a custom controller.

@romibi: I know you would have preferred this code to be merged in the core, but I still gave you credits in the License file (AGPL). I added my name to your pseudo in the 'author' field of the extension metadata so people know they can ping me if the extension breaks. Also the main difference with your version is a small delay before the icon appears in the button (since I add it in JS and I have to wait for the url variables to be available). It could be improved by adding a hook to the extension system to customize the bar directly from the backend, but it means a lot more work that I cannot pass. Hope it's fine for you!

### DIFF
--- a/xExtension-QuickCollapse/i18n/en/gen.php
+++ b/xExtension-QuickCollapse/i18n/en/gen.php
@@ -1,0 +1,7 @@
+<?php
+
+return [
+	'js' => [
+		'toggle_collapse' => 'Toggle collapse',
+	],
+];
--- a/extensions.json
+++ b/extensions.json
@@ -184,6 +184,15 @@
       "author": "Alexis Degrugillier",
       "url": "https://github.com/aledeg/FreshRSS-extensions",
       "type": "gh-subdirectory"
+    },
+    {
+      "name": "Quick Collapse",
+      "description": "Improve the sharing by email system.",
+      "description": "Quickly change from folded to unfolded articles",
+      "version": 0.1,
+      "author": "romibi and Marien Fressinaud",
+      "url": "https://github.com/FreshRSS/Extensions",
+      "type": "gh-subdirectory"
     }
   ]
 }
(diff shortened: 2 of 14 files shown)


romibi:
Sorry, I didn't find time/energy to look at it this weekend …
I'll try to have a look at it next weekend …
Thank you for doing this for me 😄

romibi:
So I found time to test it…
If the button appears, it works as intended.

But my FreshRSS installation runs on a Rapsberry Pi 2. So because of that the page load times are a bit higher than optimal. Half the times the button doesn't appear.
So I'd prefer it if the button wouldn't get loaded via JS but added directly via php …
(Or at least get added more consistantly)

For the moment I'll switch back to my branch … maybe you or me finds time to look at it more and fix this issue at a later date …

marienfressinaud:
I'll take time to fix the issue, don't worry :)

SuperSandro2000:
Did you do testing if the js lags severely when you have a very large rss feed and loaded 100 or 200 articles?

marienfressinaud:
Could you test again with the new code? I made the scripts to load once the dom is loaded so it should work, but the button might take time to appear. If it's too long for you, I'll take time to do it in PHP

xExtension-QuickCollapse/i18n/en/gen.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+return [
+	'js' => [
+		'toggle_collapse' => 'Toggle collapse',
```
Frenzie, line 5: Seems a bit unclear, although I suppose you'd be well aware what the extension does from the description. ^_^

romibi:
Sorry for not answering for so long.
I only got around to install the updated extension now …
First impression seems fine …
If I notice anything that needs improving I'll write again.

romibi:
Update: No Issues found yet.
Can be merged in my opinion.

marienfressinaud:
Thanks for the feedback!

PR #77, opened 2020-04-05
Allow Apache-friendly fine-tuning

cedric-dufour:
Hello,

This PR adds additional parameters to the `ImageProxy` extension such as to allow easy proxying via Apache `mod_rewrite` (which: 1. needs a hard-coded _absolute path_ ; 2. can not URL-decode passed URL)

See the updated `README.md` for further information (and corresponding Apache configuration example).

I've taken care to remain backward-compatible with existing setups; in particular, the (former) `force` parameter will automatically be migrated to the (new) `scheme_https` counterpart.

Best,

Cédric

### DIFF
--- a/xExtension-ImageProxy/README.md
+++ b/xExtension-ImageProxy/README.md
@@ -59,7 +59,11 @@ RewriteRule ^ http://%1 [QSA,P,L]
   Satisfy any
 </Location>
 # CRITICAL: Do NOT allow access to local resources!!!
-<LocationMatch "^/proxy/.*(localhost|127\.0\.0\.1|::1|local\.domain)">
+#  - (any) IPv4
+#  - (any) IPv6
+#  - localhost
+#  - local.domain (e.g. example.org)
+<LocationMatch "^/proxy/https?:/+([0-9]{1,3}(\.[0-9]{1,3}){3}|([0-9a-zA-Z]{0,4}:?)?(:[0-9a-zA-Z]{1,4}:){0,6}([.:][0-9a-zA-Z]{1,4}){1,7}|[^/]*(localhost|local\.domain|example\.org))(/|$)">
   Require all denied
 </LocationMatch>
 ```
--- a/xExtension-ImageProxy/README.md
+++ b/xExtension-ImageProxy/README.md
@@ -12,7 +12,8 @@ To use it, upload this entire directory to the FreshRSS `./extensions` directory
 
 * `scheme_https` (default: `0`): whether to proxy HTTPS resources
 
-* `scheme_default` (default: `http`): which scheme to use for resources that do not include one (if set to `-`, those will not be proxied)
+* `scheme_default` (default: `auto`): which scheme to use for resources that do not include one; if set to `-`, those will not be proxied;
+  if set along `scheme_include`, the scheme included in the URL will either be `auto`-matically derived from your current connection or the one explicitly specified
 
 * `scheme_include` (default: `0`): whether to include the scheme - `http*://` - in the proxied URL
 
--- a/xExtension-ImageProxy/README.md
+++ b/xExtension-ImageProxy/README.md
@@ -39,7 +39,7 @@ In order to use Apache `mod_rewrite`, you will need to set the following setting
 Along the following Apache configuration for the `www.example.org` virtual host:
 
 ```
-  # WARNING: Multiple '/' are trimmed to a single one!
+  # WARNING: Multiple '/' in %{REQUEST_URI} are internally trimmed to a single one!
   RewriteCond %{REQUEST_URI} ^/proxy/https:/+(.*)$
   RewriteRule ^ https://%1 [QSA,P,L]
   RewriteCond %{REQUEST_URI} ^/proxy/http:/+(.*)$
--- a/xExtension-ImageProxy/README.md
+++ b/xExtension-ImageProxy/README.md
@@ -28,7 +28,7 @@ The source code for the images.weserv.nl proxy can be found at [github.com/andri
 
 ### Apache configuration
 
-In order to use Apache `mod_rewrite`, you will need to set the following settings:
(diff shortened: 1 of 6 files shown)


xExtension-ImageProxy/README.md
```diff
@@ -4,10 +4,57 @@ This FreshRSS extension allows you to get rid of insecure content warnings or di
...
 
+By ticking the `scheme_https` checkbox, you can also force the use of the proxy, even for images coming through an encrypted channel. This makes the server that hosts your FreshRSS instance the only point of entry for images, preventing your client from connecting directly to the RSS sources to recover them (which could be a privacy concern in extreme cases).
+
 The source code for the images.weserv.nl proxy can be found at [github.com/andrieslouw/imagesweserv](https://github.com/andrieslouw/imagesweserv), but of course other methods are available. For example, in Apache you could [use `mod_rewrite` to set up a simple proxy](https://httpd.apache.org/docs/2.2/rewrite/proxy.html) and similar methods are available in nginx and lighttpd. Alternatively you could use a simple PHP script, [along these lines](https://github.com/Alexxz/Simple-php-proxy-script). Keep in mind that too simple a proxy could introduce security risks, which is why the default proxy processes the images.
```
Frenzie, line 27:
```suggestion
The source code for the images.weserv.nl proxy can be found at [github.com/andrieslouw/imagesweserv](https://github.com/andrieslouw/imagesweserv), but of course other methods are available. For example, in Apache you could [use `mod_rewrite` to set up a simple proxy](#apache-configuration) and similar methods are available in nginx and lighttpd. Alternatively you could use a simple PHP script, [along these lines](https://github.com/Alexxz/Simple-php-proxy-script). Keep in mind that too simple a proxy could introduce security risks, which is why the default proxy processes the images.
```

xExtension-ImageProxy/README.md
```diff
@@ -4,10 +4,57 @@ This FreshRSS extension allows you to get rid of insecure content warnings or di
...
+Along the following Apache configuration for the `www.example.org` virtual host:
+
+```
+  # WARNING: Multiple '/' are trimmed to a single one!
+  RewriteCond %{REQUEST_URI} ^/proxy/https:/+(.*)$
+  RewriteRule ^ https://%1 [QSA,P,L]
+  RewriteCond %{REQUEST_URI} ^/proxy/http:/+(.*)$
+  RewriteRule ^ http://%1 [QSA,P,L]
+  <Location "/proxy/">
+    # CRITICAL: Do NOT leave your proxy opened to everyone!!!
```
Frenzie, line 48: Grammar nit
```suggestion
    # CRITICAL: Do NOT leave your proxy open to everyone!!!
```

xExtension-ImageProxy/extension.php
```diff
@@ -1,45 +1,95 @@
...
+			}
+		}
+		else if ($scheme === 'https') {
+			if (!FreshRSS_Context::$user_conf->image_proxy_scheme_https) return $url;
+			if (!FreshRSS_Context::$user_conf->image_proxy_scheme_include) {
+				$url = substr($url, 8);  // https://
+			}
+		}
+		else if (empty($scheme)) {
+			if (substr(FreshRSS_Context::$user_conf->image_proxy_scheme_default,0, 4) !== 'http') return $url;
```
Frenzie, line 81: Does this break with the purposeful behavior described in the comment on line 36 and executed on line 39? I'd suggest making it an option if you consider it undesirable. I didn't stick it in just for fun. ;-)

See https://www.paulirish.com/2010/the-protocol-relative-url/ for what it's talking about. A URL that looks like `//websbestgifs.net/kittyonadolphin.gif`.

cedric-dufour: You're right! Sorry, I did not mean to break this behavior. I believe https://github.com/FreshRSS/Extensions/pull/77/commits/cc24497c5c42b79131b770e9c0b755fc7c906d18 fixes this and maintains backward compatibility.

Alkarex:
I trust you are the best to review this PR, @Frenzie :-)
In principle, that sounds good. One thing though - but I believe this is something to configure in Apache and not in this FreshRSS extension - is that it might allow an attacker to trigger some requests that would otherwise be disallowed. For instance querying a service running on localhost. So we should keep an eye on those aspects.

Frenzie:
You mean whether the sample Apache configuration is safe enough?

Alkarex:
Probably safe enough for thie PR @Frenzie , but could potentially be improved: I have only quickly looked at it, but it seems to filter on incoming IP (which is good) but not on destination (it could be good to exclude e.g. localhost in IPv4 and IPv6).

cedric-dufour:
> on destination (it could be good to exclude e.g. localhost in IPv4 and IPv6).

Indeed! I believe https://github.com/FreshRSS/Extensions/pull/77/commits/5881c20661120d7ac375e7a7e64f19b0861d570f addresses this issue.

Hopefully the `CRITICAL: ...!!!` stanza will raise awaress that care must be taken.

xExtension-ImageProxy/README.md
```diff
@@ -4,10 +4,62 @@ This FreshRSS extension allows you to get rid of insecure content warnings or di
...
+
+* `scheme_include` = **1**
+
+* `url_encode` = **0**
+
+Along the following Apache configuration for the `www.example.org` virtual host:
 
-By ticking the dedicated checkbox, you can also force the use of the proxy, even for images coming through an encrypted channel. This makes the server that hosts your FreshRSS instance the only point of entry for images, preventing your client from connecting directly to the RSS sources to recover them (which could be a privacy concern in extreme cases).
+```
+  # WARNING: Multiple '/' in %{REQUEST_URI} are internally trimmed to a single one!
```
Frenzie, line 43: Btw, the two spaces in front of everything seem a touch random?

cedric-dufour: Sorry. Copy-paste from my test config (fixed) :-)

Frenzie:
Thanks!

PR #78, opened 2020-04-10
restore php5.6 compatibility

Purexo:
old version of php doesn't support visibility on class const

Sorry, my server have an old version of debian and php (debian 8 - php5.6). I think i'm not alone so I purpose this small change (Tested live on my production server as very a professional developer^^).

Do whatever you want with this proposal.

Best regards


Alkarex:
Ah, we should also update `xExtension-ImageProxy/metadata.json`

Issue #72, opened 2020-01-28
Make a reference to greader "redate" extension ?


javerous:
Hi,

Following this ticket https://github.com/FreshRSS/FreshRSS/issues/2759 and this PR https://github.com/FreshRSS/FreshRSS/pull/2773, I created this extension: https://github.com/javerous/freshrss-greader-redate

At first, I wanted to keep it private, but I thought it can perhaps be used by someone else. Do you see any interest to make a reference to it on the community page (via _extensions.json_ file) ?

I'm not sure it has the quality grade expected to be on this page, and I'm not sure it worth putting it there at all… Then feel free to reject ;)

As a very short remember, the idea is to change fetching date by publishing date in greader API for clients like Reeder or EasyRSS which use only crawling date. It requires master / 1.16 version of FreshRSS to work.

[EDIT] The point is it's mainly a workaround extension. It's possible that those clients fix how they are using dates from greader at some point, making this extension _de-facto_ obsolete.

Frenzie:
There's plenty of stuff that'll become obsolete for some reason or other. I don't see that as an issue. ;-)

Alkarex:
@javerous Please send a PR with a reference to your extension

javerous:
I created the PR https://github.com/FreshRSS/Extensions/pull/79

PR #79, opened 2020-04-19
Add "GReader Redate" extension.

javerous:
Add the Greader Redate extension as suggested there https://github.com/FreshRSS/Extensions/issues/72

It's related to the (long) discussion there https://github.com/FreshRSS/FreshRSS/issues/2759

I know that @Alkarex fixed the problem in EasyRSS, but for now I would prefer to keep it in my extension, so the workaround is still possible for those who still use the old version. Perhaps the description can specify a version for EasyRSS ? But there are different forks, if I understand well, so… :s

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -163,9 +163,9 @@
     },
     {
       "name": "GReader Redate",
-      "description": "Workaround a dating problem on GReader API with Reeder and EasyRSS.",
+      "description": "Works around a dating problem on GReader API with Reeder.",
       "version": 1.0,
-      "author": "Julien Avérous",
+      "author": "Julien AvÃ©rous",
       "url": "https://github.com/javerous/freshrss-greader-redate",
       "type": "gh-subdirectory"
     }
--- a/extensions.json
+++ b/extensions.json
@@ -160,6 +160,14 @@
       "author": "Marien Fressinaud",
       "url": "https://github.com/FreshRSS/Extensions",
       "type": "gh-subdirectory"
+    },
+    {
+      "name": "GReader Redate",
+      "description": "Workaround a dating problem on GReader API with Reeder and EasyRSS.",
+      "version": 1.0,
+      "author": "Julien Avérous",
+      "url": "https://github.com/javerous/freshrss-greader-redate",
+      "type": "gh-subdirectory"
     }
   ]
 }
--- a/extensions.json
+++ b/extensions.json
@@ -163,7 +163,7 @@
     },
     {
       "name": "GReader Redate",
-      "description": "Works around a dating problem on GReader API with Reeder.",
+      "description": "Works around a dating problem on GReader API with Reeder and EasyRSS (prior to 0.7.5).",
       "version": 1.0,
       "author": "Julien Avérous",
       "url": "https://github.com/javerous/freshrss-greader-redate",


Frenzie:
> But there are different forks, if I understand well, so… :s

Not really, there's the original abandoned one and the maintained (mainly against bit rot) fork. ;-) Unless you know more?

extensions.json
```diff
@@ -160,6 +160,14 @@
       "author": "Marien Fressinaud",
       "url": "https://github.com/FreshRSS/Extensions",
       "type": "gh-subdirectory"
+    },
+    {
+      "name": "GReader Redate",
+      "description": "Workaround a dating problem on GReader API with Reeder and EasyRSS.",
```
Frenzie, line 166:
```suggestion
      "description": "Works around a dating problem on GReader API with Reeder.",
```

Or "Workaround for".

And I'd indeed leave off the EasyRSS, or at least put in something like "prior to 1.7.0".

javerous:
@Frenzie Not specifically, I just searched on Google, and saw different GitHub results (including the one from @Alkarex). I'm not sure which one is the "new one".

Anyway: @Alkarex if your version is the most maintained (at least, it's the one I tested when I did my tests with the GReader API), I can add the version in the description of extensions.json to make it more explicit.

Something like `Workaround a dating problem on GReader API with Reeder and EasyRSS (<= 0.7.5).` (the fix is still not merged).

Frenzie:
> Not specifically, I just searched on Google, and saw different GitHub results (including the one from @Alkarex). I'm not sure which one is the "new one".

The others are just random forks in the GH sense (afaik/afaict), not forks in the LibreOffice sense.

Alkarex:
Regarding EasyRSS, I am waiting to have version 7.0.4 published in F-Droid (mainly a version upgrade of the whole tool stack so that it can compile in current Arduino IDE and current Android), and then merge the PR and issue a new version. This way we have a recent reference to use as a baseline

javerous:
@Alkarex Do you know which version it will be ? For now I guessed it will be 7.0.5, but perhaps you want to jump to a new minor version, like 7.1.0 ?

Alkarex:
Probably 0.7.5

javerous:
There is a problem with the `extensions.json` no ? I don't see the list anymore in FreshRSS, and I have this log :
```
2020-06-14 12:14:27
Failed to convert extension file list
```

I don't think it's the change I did, but can it be related to
```
{
      "name": "LatexSupport",
      "description": "Add support for LaTeX notation rendering",
      "version": "0.1.2",
      "author": "Alexis Degrugillier",
      "url": "https://github.com/aledeg/FreshRSS-extensions",
    }
```
(the trailing `,` at the `url` + perhaps the `type` which is missing ?)

Frenzie:
@javerous JSON rather annoyingly doesn't support simple comments or trailing commas. See https://json5.org/ for what we probably all wish JSON were actually like.

So yes, get rid of that comma. ;-)

Btw, DDG has a built-in JSON validator if you type [`JSON validator`](https://duckduckgo.com/?q=json+validator).

Frenzie:
Oh pardon, I mean add that `"type": "gh-subdirectory"` since it looks like it was removed by accident, not on purpose. But same difference.

PR #83, opened 2020-06-16
Feature/automated generation

aledeg:
Instead of manually maintaining the file, there is now a tool that grab
all repositories defined in the repositories.json file and browse through
all repositories to find extensions.
To do that, I've added a container that runs only php-cli to trigger said
tool.

The next step will be to trigger that tool automatically on a regular basis
and to update the current repository when there are changes.

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -1,0 +1,3 @@
+FROM php:8.1-cli-alpine
+
+RUN apk add --no-cache git
--- a/generate.php
+++ b/generate.php
@@ -1,0 +1,61 @@
+<?php
+
+const VERSION = 0.1;
+const TYPE_GIT = 'git';
+
+$extensions = [];
+
+// --------------------------------------------------------------- //
+// Parse the repositories.json file to extract extension locations //
+// --------------------------------------------------------------- //
+$repositories = json_decode(file_get_contents('repositories.json'), true);
+if (JSON_ERROR_NONE !== json_last_error()) {
+	echo 'The repositories.json file is not a valid JSON file.', PHP_EOL;
+	die;
+}
+
+$gitRepositories = [];
+foreach ($repositories as $repository) {
+	if (null === $url = $repository['url'] ?? null) {
+		continue;
+	}
+	if (TYPE_GIT === $repository['type'] ?? null) {
+		$gitRepositories[] = $url;
+	}
+}
+
+// ---------------------------------------- //
+// Clone git repository to extract metadata //
+// ---------------------------------------- //
+foreach ($gitRepositories as $key => $gitRepository) {
+	echo 'Processing ', $gitRepository, ' repository', PHP_EOL;
+	exec("git clone --quiet --single-branch --depth 1 --no-tags {$gitRepository} /tmp/extensions/{$key}");
+
+	unset($metadataFiles);
+	exec("find /tmp/extensions/{$key} -iname metadata.json", $metadataFiles);
+	foreach ($metadataFiles as $metadataFile) {
+		$metadata = json_decode(file_get_contents($metadataFile), true);
+		if (JSON_ERROR_NONE !== json_last_error()) {
+			continue;
+		}
+		$metadata['url'] = $gitRepository;
+		$metadata['method'] = TYPE_GIT;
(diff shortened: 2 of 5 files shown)


Docker/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM php:7.4-cli
+
+RUN apt-get update && apt-get install -y --no-install-recommends git && apt-get purge -y --auto-remove
```
Alkarex, line 3: I would suggest dropping this Dockerfile, using `php:8-cli` directly instead in the Makefile, and take advantage of git on the client

Alkarex:
I somehow did not see this PR until now, sorry!
Could you resolve the conflicts?

Docker/Dockerfile
Alkarex: Ah sorry, that is not possible because you use git inside Docker

Alkarex: But maybe consider `php:8-cli-alpine` then, which is much smaller and with a much lighter build

Alkarex: Example https://github.com/FreshRSS/FreshRSS/blob/896f9d33637ec510937bc13a34f98f669795ab49/Docker/Dockerfile-Alpine#L1-L9

aledeg:
I will as soon as I resume my work on that. I'll keep you posted!

Docker/Dockerfile
aledeg: Done!

generate.php
```diff
@@ -0,0 +1,63 @@
...
+        $gitRepositories[] = $url;
+    }
+}
+
+// ---------------------------------------- //
+// Clone git repository to extract metadata //
+// ---------------------------------------- //
+foreach ($gitRepositories as $key => $gitRepository) {
+    echo 'Processing ', $gitRepository, ' repository', PHP_EOL;
+    exec("git clone --quiet --single-branch --no-tags {$gitRepository} /tmp/extensions/{$key}");
```
Alkarex, line 34:
```suggestion
    exec("git clone --quiet --depth 1 --no-tags {$gitRepository} /tmp/extensions/{$key}");
```

No need to clone the whole history for that

Alkarex:
It would be nice to keep downloaded repositories in a cache to avoid too much resource consumption.
And if the cache already exists, then use a a git fetch and git reset instead of git clone.

generate.php
aledeg: I've added the `depth` option and kept the `single-branch` also.

aledeg:
> It would be nice to keep downloaded repositories in a cache to avoid too much resource consumption.
> And if the cache already exists, then use a a git fetch and git reset instead of git clone.

I think it would be really hard for no real gain.
My goal is to trigger the generation tool once a day automatically with github actions. I am pretty sure I cannot keep cache between executions. The idea is to have the file always up-to-date with a delay of 24 hours maximum and to avoid manual work.

Frenzie:
With the exception of the two framagit repos, by being on GitHub they're already in cache for all intents and purposes.

@aledeg Is this focus on master and --no-tags correct for all of the extensions? One may well update the json long before tagging a release.

aledeg:
@Frenzie You're right. Do you have any suggestion on how I could address that?
Do we keep a policy in the file description?
Do we use a single rule for all repositories?

Frenzie:
A single rule of tags if there are any, otherwise json might do the trick. But indeed, manual policy is probably better/safer. Not quite sure.

aledeg:
I'll look into that!

Alkarex:
A little conflict :-)

aledeg:
I've solved the conflict. I'll try to find a way to address @Frenzie 's concern about the tagging issue. I have some ideas but I have to dig a little bit.

aledeg:
I need also to dig into github actions to see how I can build the list automatically. That will be fun, I am looking forward to it.

aledeg:
Oh! shoot.  I was not ready :/
I'll make another branch.

Alkarex:
Ah, I thought you wanted the CI in another PR :-)

aledeg:
Don't worry. I'll make the changes in another PR :)

PR #93, opened 2020-11-14
Add RSS-Bridge extension

DevonHess:
Added an extension to run any URL you add through [RSS-Bridge](https://github.com/rss-bridge/rss-bridge).

RSS-Bridge is a project that creates RSS feeds for websites that don't have one. This plugin basically allows FreshRSS to support feeds from hundreds of new websites.

It's a very simple extension. If RSS-Bridge doesn't support the feed, FreshRSS will try to add your URL normally, but if it does support the feed, it'll add an RSS-Bridge feed instead.

This is my first ever pull request. Let me know if I did anything wrong/stupid. Thanks.


Alkarex:
Nice, thanks 👍

Issue #94, opened 2020-11-15
ImageProxy cannot handle html encode correctly.


Victrid:
When adding origin url with special characters like "&", ImageProxy will parse it into &amp; , which is not as expected.

It's can be reproduced setting proxy url to `https://example.com/proxy?token=abcdef&url=`

Frenzie:
I'm afraid you're either mistaken or something about your problem description is slightly incomplete. `&amp;url` is definitely expected and desired since `&url` is invalid HTML.

If you input:
```
https://example.com/proxy?token=abcdef&url=
```

What you need to see in HTML is:
```
<img src="https://example.com/proxy?token=abcdef&amp;url="
```

Victrid:
Umm, I'm confused.

When using 
```
https://example.com/proxy?token=abcdef&url=
````
I get tags like 
```
<img src="https://example.com/proxy?token=abcdef&amp;amp;url=...">
```
and my nginx server error log says
```
invalid URL prefix in "", client: ..., server: example.com, request: "GET /proxy?token=abcdef&amp;url=https://contoso.com/some.jpg HTTP/1.1", host: "rexample.com"
```
the browser should not send servers with escaped urls.

It seems that the `&amp;` is escaped twice. 

Please see this:

https://github.com/FreshRSS/Extensions/blob/5ddbae989b66dca56bbab1c96db72bcfd3deb661/xExtension-ImageProxy/extension.php#L120

Maybe php did the escape job automatically. As [this Stack Overflow page](https://stackoverflow.com/questions/19545373/how-to-stop-php-dom-from-auto-escaping-character-references) described, `setAttribute` will automatically escape the contents, which could cause the same problem.

Frenzie:
> I get tags like
>
>`<img src="https://example.com/proxy?token=abcdef&amp;amp;url=...">`


Well, there you go then. That's quite different from `&` is turned into `&amp;`. ;-)

Please fix it like this instead:

```diff
diff --git a/xExtension-ImageProxy/extension.php b/xExtension-ImageProxy/extension.php
index 504248b..2e9e3c9 100644
--- a/xExtension-ImageProxy/extension.php
+++ b/xExtension-ImageProxy/extension.php
@@ -52,7 +52,7 @@ class ImageProxyExtension extends Minz_Extension {
                $this->registerTranslates();
 
                if (Minz_Request::isPost()) {
[...]

PR #95, opened 2020-11-15
add decoding for html input (#94)

Victrid:
The problem described can be found in #94,

>When adding origin url with special characters like "&", ImageProxy will parse it into & , which is not as expected.
>
>It's can be reproduced setting proxy url to `https://example.com/proxy?token=abcdef&url=`


Frenzie (commented):
Pardon, I wrote my review in https://github.com/FreshRSS/Extensions/issues/94#issuecomment-727608469

Victrid:
Now the file is changed as suggested in #94.

Frenzie (approved):
Great, thanks! :-)

PR #103, opened 2021-01-05
Fix extension definitions

aledeg:
Before, extensions were missing the definition of mandatory methods.
At the moment, it's not a problem because the coding guidelines are not
enforced by the code. But in the future, it will break.
Now, extensions have all mandatory method definitions.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -4,7 +4,7 @@
     {
       "name": "Custom CSS",
       "description": "A FreshRSS extension which give ability to create user-specific CSS rules to apply in addition of the actual theme",
-      "version": 0.2,
+      "version": 0.3,
       "author": "Marien Fressinaud",
       "url": "https://github.com/FreshRSS/Extensions",
       "type": "gh-subdirectory"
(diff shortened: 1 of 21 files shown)


extensions.json
```diff
@@ -208,6 +208,22 @@
...
+      "version": 0.2,
+      "author": "Marien Fressinaud",
+      "url": "https://github.com/FreshRSS/Extensions",
+      "type": "gh-subdirectory"
+    },
+    {
+      "name": "Title-Wrap",
+      "description": "Applies a line-wrap to long article titles, as opposed to truncating the title when it overflows its display area.",
+      "version": 0.2,
+      "author": "₣rans de Jonge, Joris Kinable",
```
Frenzie, line 224: I don't mind the extra flourish, but how'd you even type that? :-P
```suggestion
      "author": "Frans de Jonge, Joris Kinable",
```

Frenzie (commented):
The extensions define their own versions in their metadata. This extensions.json is a file that can be downloaded by FreshRSS to see if extensions are still up to date.
https://github.com/FreshRSS/Extensions/blob/17f42e4b53b0ce865a4272befb881b73383a02d4/xExtension-QuickCollapse/metadata.json#L5

extensions.json
aledeg: Fixed

PR #108, opened 2021-01-22
Add new Explosm extension

dohseven:
Add a new [Explosm extension](https://framagit.org/dohseven/freshrss-explosm), which is able to retrieve the Explosm comic to display it directly in FreshRSS.


Frenzie:
Thanks!

PR #117, opened 2021-08-10
ImageProxy: Add nginx proxy example.

tastytea:
It took me a while to figure out how to do a caching proxy with nginx and there are open issues in FreshRSS/FreshRSS for this feature. So I think this example might be useful to others.

### DIFF
--- a/xExtension-ImageProxy/README.md
+++ b/xExtension-ImageProxy/README.md
@@ -67,3 +67,46 @@ RewriteRule ^ http://%1 [QSA,P,L]
   Require all denied
 </LocationMatch>
 ```
+
+### nginx configuration
+
+In order to use nginx's [proxy
+module](https://nginx.org/en/docs/http/ngx_http_proxy_module.html), you will
+need to set the following settings:
+
+* `proxy_url` = **https://www.example.org/proxy?key=changeme&url=**
+* `scheme_include` = **1**
+* `url_encode` = **0**
+
+Add this to your nginx config:
+
+``` nginx
+# Use 1 GiB cache with a 1 MiB memory zone (enough for ~8,000 keys).
+# Delete data that has not been accessed for 30 minutes.
+proxy_cache_path /var/cache/nginx/freshrss levels=1:2 keys_zone=freshrss:1m
+                   max_size=1G inactive=30m use_temp_path=off;
+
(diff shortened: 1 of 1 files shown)


xExtension-ImageProxy/README.md
```diff
@@ -67,3 +67,46 @@ RewriteRule ^ http://%1 [QSA,P,L]
...
+
+* `proxy_url` = **https://www.example.org/proxy?key=changeme&url=**
+* `scheme_include` = **1**
+* `url_encode` = **0**
+
+Add this to your nginx config:
+
+``` nginx
+# Use 1 GiB cache with a 1 MiB memory zone (enough for ~8,000 keys).
+# Delete data that has not been accessed for 30 minutes.
```
Frenzie, line 85: 30 minutes is for a very busy server? It doesn't matter much since it's just an example but I'd be more inclined to put in a value closer to, let's say, 36h, so I'm curious about the rationale behind 30 minutes in any case. :-)

Frenzie (approved):
Sure, lgtm.

xExtension-ImageProxy/README.md
tastytea: I didn't put too much thought in it to be honest. Theoretically images could change after publishing, but I guess that rarely happens. How about I change it to delete after 12 hours if inactive, 2 days maximum? I think most feed articles are read within a day, so I don't see the point in keeping it much longer.

Frenzie: Sure. ^_^

Frenzie:
@Alkarex Any comments?

Alkarex:
> @Alkarex Any comments?

@Frenzie I have not tried it myself, but it looks all good to me :-)

wclmgcd:
> 我花了一段时间才弄清楚如何使用 nginx 做一个缓存代理，并且在 FreshRSS/FreshRSS 中有针对此功能的未解决问题。 所以我认为这个例子可能对其他人有用。


Can you add a proxy example for caddy?
可以添加一个caddy 的代理示例吗？

tastytea:
> Can you add a proxy example for caddy? 可以添加一个caddy 的代理示例吗？

I don't know anything about caddy.

PR #125, opened 2022-01-03
StickyFeeds: Update README.md

math-GH:
Mark this extension as deprecated, because it is already implemented into FreshRSS

### DIFF
--- a/xExtension-StickyFeeds/README.md
+++ b/xExtension-StickyFeeds/README.md
@@ -1,9 +1,9 @@
-# Sticky Feeds extension (Depricated)
+# Sticky Feeds extension (Deprecated)
 
 A FreshRSS extension which set the feed aside in the main stream following the window scroll.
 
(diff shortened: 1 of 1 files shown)


xExtension-StickyFeeds/README.md
```diff
@@ -1,5 +1,9 @@
-# Sticky Feeds extension
+# Sticky Feeds extension (Depricated)
```
Frenzie, line 1: deprecated :-)

But I think all files except for this one should be removed, leaving just this README file, also stripped of at least that last line.

Frenzie: Also to be removed (you could add it to this PR if you like)
https://github.com/FreshRSS/Extensions/blob/e69e454618dd63bece0e275832ad984799f59bc3/extensions.json#L28-L35

Alkarex (approved):

PR #127, opened 2022-01-09
Add github action to build extension list

aledeg:


Frenzie (commented):
Is the current result identical btw? :-)

aledeg:
@Frenzie What do you mean?

Frenzie:
Just curious if the resulting JSON is the same.

aledeg:
Yes, the resulting JSON is the same. I mean the same as if we generate it manually with the tool that has been merged yesterday. The tool change it slightly compared with what we have previously though,
But overall, this won't change how the reader will use it. You can even check it by yourself on your instance. The list of extension is extracted from that file.

.github/workflows/generate.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    env:
+      FILE: extensions.json
+
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v2
+
+      - name: Build JSON file
+        run: php ./generate.php
```
Alkarex, line 19: Maybe make sure that the main errors will return a non-zero error code, so that the workflow will abort

https://github.com/FreshRSS/Extensions/blob/93c042a57d7f45d7b8aafc71cdffe689998152f9/generate.php#L14

aledeg: Good point. I'll add something to handle that better.

aledeg: @Alkarex I've refactor the code to leverage PHP 7.3 features. As it should run on PHP 8.1 in the container, we should be fine.

generate.php
```diff
@@ -8,10 +8,11 @@
...
 // Parse the repositories.json file to extract extension locations //
 // --------------------------------------------------------------- //
-$repositories = json_decode(file_get_contents('repositories.json'), true);
-if (JSON_ERROR_NONE !== json_last_error()) {
+try {
+	$repositories = json_decode(file_get_contents('repositories.json'), true, 512, JSON_THROW_ON_ERROR);
+} catch (Exception $exception) {
 	echo 'The repositories.json file is not a valid JSON file.', PHP_EOL;
-	die;
+	exit(1);
```
Alkarex, line 15: No need to change it, but I often use `die(1)` (or another code) for errors and `exit(0)` when there is no error - the two are synonyms

Alkarex:
It is not really a part of this PR, but a few comments, not required (feel free to ignore):

* Running `generate.php` twice leads to errors such as `fatal: destination path '/tmp/extensions/10' already exists and is not an empty directory.` Either remove the directories or prepare for `git fetch ... && git reset ...`
* `generate.php` could be made `+x` executable
* Maybe instead of `/tmp/extensions/` the working directory could be `./data/extensions/` with the corresponding `.gitignore` to reduce the risks of collisions and abandoned temp files
* The `key` in `foreach ($gitRepositories as $key => $gitRepository) {` could be a hash (e.g. sha1) of the repository URL for more stability than 0, 1, 2, 3...

.github/workflows/generate.yml
```diff
@@ -0,0 +1,37 @@
+name: Generate extension list
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
Alkarex, line 5: Maybe a random hour and minute instead of the too-usual midnight

generate.php
```diff
@@ -8,10 +8,11 @@
 // --------------------------------------------------------------- //
 // Parse the repositories.json file to extract extension locations //
 // --------------------------------------------------------------- //
-$repositories = json_decode(file_get_contents('repositories.json'), true);
```
Alkarex, line 11: Ah, something more, `git clone` hangs on a wrong URL, e.g. `https://github.com/kevinpapst/freshrss-dilbert2`

aledeg:
@Alkarex I've addressed all your concerns. It should work as intended without making a difference in the output.

extensions.json
```diff
@@ -261,18 +261,18 @@
...
             "type": "system",
-            "url": "https:\/\/github.com\/FreshRSS\/Extensions",
+            "url": "https://github.com/FreshRSS/Extensions",
             "method": "git",
             "directory": "xExtension-TTRSS_API"
         },
         {
             "name": "Title-Wrap",
-            "author": "\u20a3rans de Jonge, Joris Kinable",
+            "author": "₣rans de Jonge, Joris Kinable",
```
Alkarex, line 270: Is the ₣ on purpose, @Frenzie ? :-)

Frenzie: No, see https://github.com/FreshRSS/Extensions/pull/105

Alkarex: It is because it is not fixed in https://github.com/FreshRSS/Extensions/blob/master/xExtension-TitleWrap/metadata.json

PR #139, opened 2022-08-10
Image Cache Extension

Victrid:
This is an extension for image caching.

Different to Image Proxy plugin, this plugin will notify the cache server to download images when it receives feed entries, and it will be faster to load the picture if the cache server already has the image. While cache missed, this can also function as ImageProxy plugin does.

We also have a cloudflare worker as the example implementation of the cache server, which will utilize the cloudflare cache API so the cached images will be delivered by cloudflare CDN. It can be hosted on cloudflare for free.

Not only solving the http/https degrade problem, users can implement their own cache server or modify the example implementation to fetch images with additional headers, and different user agent, as some RSS feeds may have misconfigured policies on image hotlinking.


Frenzie (approved):
Looks cool!

Alkarex:
Thanks!

Related:
* https://github.com/FreshRSS/FreshRSS/issues/1215
* https://github.com/FreshRSS/FreshRSS/issues/3798
* https://github.com/FreshRSS/FreshRSS/discussions/4486

PR #141, opened 2022-08-17
Fix #116 (reading time duplication)

### LINKED ISSUE #116: xExtension-ReadingTime displays reading time multiple times
I installed this extension by downloading [the extension archive](https://github.com/FreshRSS/Extensions/archive/master.zip), extracting it, and moving the xExtension-ReadingTime folder to `./extensions` as instructed, and then I enabled it via the extensions configuration menu. I tried disabling and then re-enabling it, disabling the only other extension that I installed alongside it (xExtension-TitleWrap), and I tried using different themes but the problem persists no matter what.

Additional reading times ~~will continue getting tacked on pretty much every time I expand and them collapse stories as I read my feeds~~ get appended as I scroll down to load more articles.

screenshot

Edited for initially incorrectly reporting observed behavior

Siphalor:
Also removed some commented debug logging statements.

I don't really understand why the current implementation of this extension prefers object properties over local variables, but I decided not to touch it :)

### DIFF
--- a/xExtension-ReadingTime/static/readingtime.js
+++ b/xExtension-ReadingTime/static/readingtime.js
@@ -13,15 +13,19 @@
 
  init: function() {
      var flux_list = document.querySelectorAll('[id^="flux_"]');
-     // console.log(flux_list)
 
      for (var i = 0; i < flux_list.length; i++) {
-         //console.log($("div[id^='flux']")[i], "Length (words): ", reading_time.flux_words_count($("div[id^='flux']")[i]))
+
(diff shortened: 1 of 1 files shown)


Alkarex:
Ping @lapineige

Thanks @Siphalor 👍🏻

lapineige:
Thank you @Siphalor for fixing this, I did not have the time nor the skills to fix this :laughing:

> I don't really understand why the current implementation of this extension prefers object properties over local variables, but I decided not to touch it :)

Because I'm a Javascript noob that did it in a quick and dirty way as fast as I could :sweat_smile:

PR #147, opened 2022-12-21
Add TogglableMenu extension

nicofrand:


Frenzie (approved):

nicofrand:
Hu, why does the GitHub Actions Bot remove my modifications?

Frenzie:
Whoops, I forgot. You should add it to https://github.com/FreshRSS/Extensions/blob/f66efcf5feb1245607c2e3ac9ef4176b5f2801b3/repositories.json and then it'll be auto-generated. :-)

nicofrand:
Thx, I will. 
What is the point of having both an entry in the repositories.json and the extensions.json though?

Frenzie:
I'm not sure really; I suppose you could do all that with just the one file. You'll have to ask @aledeg.

My guess would be it's so that repositories.json is guaranteed to remain intact and isn't messed with by any scripts.

nicofrand:
Ah it seems adding it to repositories.json automatically adds it to extensions.json!

Frenzie:
Well yes, I thought you were asking why use two files instead of just one. extensions.json also contains the url and method, so it could also pull in/update the rest automatically and overwrite itself. You can do the same thing locally with generate.php That's all the action does. ^_^

aledeg:
> Thx, I will. What is the point of having both an entry in the repositories.json and the extensions.json though?

One of those is supposed to be deleted when the whole extension refactor is done. But it is not at the moment.

nicofrand:
Hi!

Is there something missing to see this MR merged?

Thx

PR #149, opened 2023-02-03
Add AutoTTL extension

mgnsk:
This is an extension that leverages the `feed_before_actualize` hook to skip refreshing feeds that most likely don't have any new items based on the average frequency of entries.

Screenshot 2023-02-03 at 17-47-56 Extensions · FreshRSS

Open to any suggestions.


Frenzie:
Thanks! Interesting concept. I do that sometimes, but manually. :-)

PR #151, opened 2023-03-21
Add RSS-Bridge extension

DevonHess:
Added RSS-Bridge extension to extension list and repos. Already in readme.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -288,6 +288,17 @@
             "directory": "."
         },
         {
+            "name": "RSS-Bridge",
+            "author": "Devon Hess",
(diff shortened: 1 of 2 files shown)


extensions.json
```diff
@@ -287,6 +287,17 @@
             "method": "git",
             "directory": "."
         },
+        {
+            "name": "RSS-Bridge",
+            "author": "Devon Hess",
+            "description": "Run URLs through <a href=\"https://github.com/rss-bridge/rss-bridge\">RSS-Bridge</a> detection.",
```
Alkarex, line 293: I believe it is the first example of HTML use in the description field, which we support, but with which we should be extra careful.

Issue #148, opened 2022-12-23
Reading time extension error


svemoe:
Reading time extension is not working for me.
In firefox console I get the error message `Uncaught TypeError: document.querySelector(...).childNodes[0].childNodes[2] is undefined`.
Maybe something in FreshRSS layout changed?
Anyone else have this problem?
When I switch to mobile view it's fine.

Frenzie:
At a quick glance I think that'll work as intended if you use ` document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.website span") ` but I'm not 100% sure what it used to look like.

math-GH:
@svemoe Which FreshRSS version do you use? 1.20.2?

ping @lapineige as author of this extension.

lapineige:
It doesn't work for me on 1.20 indeed… I recently updated, I had no time to fix it.
I will try to take a look in the coming days, @svemoe don't hesitate to ping me if nothing happened in let's say 2 weeks :)

Alkarex:
Ping @lapineige ;-)

lapineige:
Fixed by #156 ?

PR #156, opened 2023-04-17
Fix & improve Reading time extension

sal0max:
As reported by @svemoe in #148, the Reading time extension doesn't work any more, beginning with 1.20.
The fix would be relatively easy, by just changing the `querySelector`.

However, I'd suggest further improvement of the plugin, by not just injecting the reading time in front of the feed name, but instead adding a separate column for it.

Looks like this with the standard theme in regular view. Notice the second to last column:
grafik

And like this in mobile view:
grafik

I also tested it with other themes, without a problem.

### DIFF
--- a/xExtension-ReadingTime/static/readingtime.js
+++ b/xExtension-ReadingTime/static/readingtime.js
@@ -27,19 +27,17 @@
 
         flux_list[i].dataset.readingTime = reading_time.reading_time;
 
-         if (document.body.clientWidth <= 840) { // in mobile mode, the feed name is not visible (there is only the favicon)
-             // add the reading time right before article's title
-             // in that case, [Time] - [Title] format is used instead of a "|" (as it looks better and doesn't take much more space)
-             if ( document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.title a").textContent.substring(0,(reading_time.reading_time + 'm - ').length) != reading_time.reading_time + 'm - ' ) {
-                 document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.title a").textContent = reading_time.reading_time + 'm - ' + document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.title a").textContent;
-             }
-         } else {
-             // add the reading time just after the feed name
-             if ( document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.website").textContent.substring(1, (reading_time.reading_time + 'm|').length + 1) != reading_time.reading_time + 'm|' ) {
-                 document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.website").childNodes[0].childNodes[2].textContent = reading_time.reading_time + 'm| ' + document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list li.item.website").childNodes[0].childNodes[2].textContent;
-             }
-         }
-
+        const li = document.createElement("li");
+        li.setAttribute("class", "item date");
+        li.style.width = "min-content";
+        li.style.minWidth = "40px";
+        li.style.overflow = "hidden";
+        li.style.textAlign = "right";
+        li.style.display = "table-cell";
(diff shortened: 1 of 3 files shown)


xExtension-ReadingTime/static/readingtime.js
```diff
@@ -16,49 +16,47 @@
...
 
-     return reading_time.textContent.split(' ').length;
+    return reading_time.textContent.split(' ').length;
  },
 
  calc_read_time : function calc_read_time(wd_count, speed) {
-     reading_time.read_time = Math.round(wd_count/speed);
-     if (reading_time.read_time === 0) { reading_time.read_time = '<1'; }
-     return reading_time.read_time;
+    reading_time.read_time = Math.round(wd_count/speed);
```
Frenzie, line 57: The indentation seems to have changed to 3 spaces while it was 4. (Also applies to the brand new code above.)

Frenzie (approved):
Nothing jumps out, pinging @lapineige

xExtension-ReadingTime/static/readingtime.js
Alkarex: Consider putting an `.editorconfig` file at the root of the project to avoid such problems https://editorconfig.org as it ensures the same convention (whatever it is) throughout the project.
See e.g. https://github.com/FreshRSS/FreshRSS/blob/edge/.editorconfig

(For the record, I personally prefer tabs, to leave users the choice of how it should look like, instead of having to fight with the - less efficient in any case - 1,2,3,4,8 space conventions)

Alkarex: Ah, actually if it project is being worked on from this repo, then we already do have a https://github.com/FreshRSS/Extensions/blob/master/.editorconfig
So please make sure your text editor is using it: https://editorconfig.org/#pre-installed

sal0max: Alright, good hint!

So, the file originally was indented with spaces. And the only setting for `*.js` files in the `.editorconfig` is to use tabs. Should I do another commit, re-indenting everything, or would this - for the sake of better traceability - better be handled in a separate pull request?

Frenzie: Yes, better to leave unaffected lines alone unless touched or to do it separately, otherwise git blame is a pain.

xExtension-ReadingTime/static/readingtime.js
```diff
@@ -27,19 +27,17 @@
...
+        li.setAttribute("class", "item date");
+        li.style.width = "min-content";
+        li.style.minWidth = "40px";
+        li.style.overflow = "hidden";
+        li.style.textAlign = "right";
+        li.style.display = "table-cell";
+        li.textContent = reading_time.reading_time + '\u2009m';
+ 
+        const ul = document.querySelector("#" + reading_time.flux.id + " ul.horizontal-list");
+        ul.insertBefore(li, ul.children[ul.children.length - 1]);
```
Frenzie, line 40: This section seems to be at 3 spaces

sal0max: I count 8 blanks in each line (<4><4>).
The lines above it are indented <1><4><4>.

That's the root of evil: https://github.com/sal0max/Extensions/blob/readingtime-tablecolumn/xExtension-ReadingTime/static/readingtime.js#L14
The `init: function() {` is indented with only one space, making everything inside it chaotic.

I've pushed another commit, adding that single space to the new code, making it uniform.
However, one should probably soon fix this in a new pull request, after this one.

lapineige:
This sounds **a lot** better, thank you for improving this 🙂

Do you agree if I had this change in my own repository, and how do I credit you ? :)

Alkarex:
Can we merge here so, @lapineige ?

sal0max:
Glad you like it, @lapineige!
No need to credit. Just take the changes however you want.

lapineige:
@sal0max then I will link to this PR in the commit, to keep track of your improvements ;)


> Can we merge here so, @lapineige ?

Yes you can 🙂

lapineige (approved):

PR #162, opened 2023-07-18
Update ThreePanesView version in extensions.json

nicofrand:


Frenzie:
If you tag it I think the bot should do it automatically.

PR #168, opened 2023-08-31
added showFeedID extension

math-GH:
inspired by the question by @cocowash in https://github.com/FreshRSS/FreshRSS/issues/5607 

I created a very simple extension, that adds a button on the subscription management page that can display the feed IDs


math-GH:
it looks like this: 
grafik

PR #169, opened 2023-09-04
added linters and fixed the issues

math-GH:
similar to the main FreshRSS this PR implements linters. The findings are fixed.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,29 @@
+{
+	"env": {
+		"browser": true
+	},
+	"extends": [
+		"eslint:recommended",
+		"standard"
+	],
+	"rules": {
+		"camelcase": "off",
+		"comma-dangle": ["warn", {
+			"arrays": "always-multiline",
+			"objects": "always-multiline"
+		}],
+		"eqeqeq": "off",
+		"indent": ["warn", "tab", { "SwitchCase": 1 }],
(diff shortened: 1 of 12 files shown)


Frenzie (approved):
Fine by me in principle, but just making sure you tested them all? Changing var to let or const can easily break things due to (lack of) hoisting.

math-GH:
I checked as far as I understand:
- Sticky Feeds extension is deprecated (because the functionality is already in FreshRSS implemented). No errors/warnings in the browser console
- xExtension-ReadingTime works fine
- xExtension-QuickCollapse: I do not understand the functionality (see #170). No errors/warnings in the browser console

math-GH:
xExtension-QuickCollapse is fixed now with the previous commit. I do not understand why `const` break it while the value will not be changed. It does not matter. Now it works fine again

All 3 affected extensions are tested and work fine

xExtension-QuickCollapse/metadata.json
```diff
@@ -2,7 +2,7 @@
 	"name": "Quick Collapse",
 	"author": "romibi and Marien Fressinaud",
 	"description": "Quickly change from folded to unfolded articles",
-	"version": 0.1,
+	"version": 0.1.1,
```
Alkarex, line 5: https://github.com/FreshRSS/Extensions/pull/176

PR #171, opened 2023-09-07
Update ReadingTime Extension for Safari Support

parispjones:
Change `li.style.width` from `min-content` to hardcoded `40px`.

Apparently Safari doesn't support `min-content` so the reading time is not shown in Safari.

I figure 40px is plenty of room for most reading times, but feel free to change it up.

Reference: https://stackoverflow.com/a/7085104


Alkarex:
Ping @sal0max

Frenzie:
If you follow the link to the fiddle in the link you'll note `min-width` does seem to work on tables in modern Safari, although given the diff is about `min-content` it's less clear how relevant that is regardless.

sal0max:
The Stack Overflow link is about Safai not supporting the `min-width` attribute.
Your PR is changing the `min-content` value of the `width` attribute, @parispjones 

- Nevertheless, `width = "min-content"` indeed causes problems with Safari not showing the column, probably because **neither** the mentioned attributes/values are properly supported.
- In Chrome, it does show up, but is way too wide.

So, yeah.. we should change something, here. I testes some combinations, and `li.style.width = "40px";` seems sufficient in all cases. Agreed.
To keep the code clean, we should remove `li.style.minWidth = "40px";`, however.

parispjones:
@sal0max You are correct. Sorry for the poor initial phrasing.
I removed minWidth so we should be good now.

Alkarex:
Please also bump the version number https://github.com/FreshRSS/Extensions/blob/master/xExtension-ReadingTime/metadata.json

parispjones:
Done!

Alkarex:
Arg, I should have merged this PR before https://github.com/FreshRSS/Extensions/pull/169
Could you try to re-apply your changes here @math-GH to fix the conflicts?

Alkarex:
Ping @math-GH

math-GH:
> Ping @math-GH

Done. Thanks for the reminder.

PR #180, opened 2023-11-17
Add 'Colorful List' Extension

shinemoon:
Target is to use different color for different feeds in list view, which can help me to quick sort /distinguish different sources by a quick glance.

snapshot


xExtension-ColorfulList/static/script.js
```diff
@@ -0,0 +1,40 @@
+$(document).ready(function(){
```
Frenzie, line 1: This probably isn't necessary when you append a script to the end of the document so I'd remove it completely (after testing), but fyi that's jQuery for:

```suggestion
document.addEventListener('DOMContentLoaded', function(){
```

xExtension-ColorfulList/static/script.js
```diff
@@ -0,0 +1,40 @@
...
+        //observer.disconnect();
+    };
+    monitorEntry(colorize);
+});
+
+
+function colorize(entries){
+    let entry = document.querySelectorAll('.flux_header');
+    entry.forEach((e,i)=>{
+        let cl = stringToColour(e.querySelector('.website').textContent)+'12';
```
Frenzie, line 23: Incidentally, looking at all this I'd be more inclined to think in this direction:

```js
let css = [...document.querySelectorAll('.item.feed')].map(el => {
 let id = el.id.slice(2)
 return `.flux_header[data-feed=${id}] {
  background-color: ${stringToColor(el.textContent)};
 }`;
}.join(''); // Joining all the strings prevents creating a bigger and bigger string along the way.
// append CSS to page
```

Though when I phrase it like that it might make more sense to do that in PHP instead. :-)

Frenzie: PS Unlike the jQuery thing that's not something to change before merging, just some general thoughts.

shinemoon: Thanks @Frenzie ! And I might be a bit concerned to touch php part as actually I am not really familiar with php 😢 Hence maybe it is 'safer' for me to make such patch-like js mocking if as you mentioned not kind of serious concern there?😆

Frenzie: Regardless of JS or PHP, I think the approach I sketched is simpler because you calculate once without needing a MutationObserver or any of that. ;-)

But I just realized the security policy may annoyingly disallow what I was proposing.
```
let stylesheet = document.createElement('style');
stylesheet.innerText = css;
document.head.appendChild(stylesheet);
```

xExtension-ColorfulList/static/script.js
```diff
@@ -0,0 +1,40 @@
...
+    });
+};
+
+
+const stringToColour = (str) => {
+  let hash = 0;
+  str.split('').forEach(char => {
+    hash = char.charCodeAt(0) + ((hash << 5) - hash)
+  })
+  let colour = '#'
```
Frenzie, line 34: PS This is a touch confusing given that in CSS it's called `color`.

Frenzie (approved):
Anyway, lgtm

xExtension-ColorfulList/static/script.js
shinemoon: > Regardless of JS or PHP, I think the approach I sketched is simpler because you calculate once without needing a MutationObserver or any of that. ;-)
> 
> But I just realized the security policy may annoyingly disallow what I was proposing.
> 
> ```js
> let stylesheet = document.createElement('style');
> stylesheet.innerText = css;
> document.head.appendChild(stylesheet);
> ```

OK, but in case following proposal above, my concern is, once we remove MutationObserver , sounds the pre-created css style will only cover the first page's list ,i.e. the style loaded later with different feed source will then not be covered by those pre(and one-shot) css styles? 

Or, in that sense, I need to generate those css styles based on fetch all feeds info in advance but not on the fly during the page load? Of course, it's doable, but, somehow it's not so 'lazy' and straightforward like now...Maybe I wrongly read your suggestion as amateur....thanks again for your time and patient on reviewing this :)

,

shinemoon:
> Anyway, lgtm

Thanks, @Frenzie , just noticed the latest reply:)

shinemoon:
> > Anyway, lgtm
> 
> Thanks, @Frenzie , just noticed the latest reply:)

... Oops.. @Frenzie  wro

xExtension-ColorfulList/static/script.js
shinemoon: Updated. Thx!

xExtension-ColorfulList/static/script.js
Frenzie: `document.querySelectorAll('.item.feed')` is the menu, or in other words every single feed there is.

shinemoon: Yes, you are right -> I wrongly thought it's grabbing from the post list panel! thanks!

Issue #183, opened 2023-11-22
xExtension-ColorfulList not working when only viewing unread messages


MarcelVersteeg:
When I view the unread messages in FreshRSS and have the ColorfulList extension enabled, the messages are not colored, like they are when viewing all messages.

Alkarex:
Ping @shinemoon

shinemoon:
@MarcelVersteeg it sounds weird, while ColorfulList supposes to 'colorize' each '.flux_header' despites of read or unread view and actually I double checked and tried also in my side by switch between 'read/unread' view there... it works. 

So is it possible you can inspect the ul item in browser to see if the item like below , i.e. background style well applied in your side: 

`<ul class="horizontal-list flux_header websitefull" style="background: rgba(62, 9, 19, 0.07);">`

MarcelVersteeg:
Hi @shinemoon 

The matter seems to be a little more complicated, but let me first answer your question about the theme: I'm using Alternative Dark.

When I opened my FreshRSS this morning and immediately went to the "Unread" messages (by clicking this button: 
![image](https://github.com/FreshRSS/Extensions/assets/72593247/35be2944-5d74-4923-93f4-6dc636c87e2f)
), the list was colored as it was when the complete list was shown.

Then I execeuted the following steps:

1. At the bottom of the list, click "Mark all as read" (this returned me to the complete list of all messages and they were correctly colored)
2. Marked some messages as "Unread" (by clicking the envelope in front of the message)
3. Open the "Unread" messages again (this shows the list of the messages marked as "Unread", but there is no coloring).

In that situation I also checked to see if the style was applied to the item and it wasn't.

As a side question: what colors are being used for the origins? Can I configure that somewhere?

shinemoon:
Hi, @MarcelVersteeg ,

Thanks, I tried the procedure you addressed, yes, it can be reproduced in my side, and this kind of issue is caused by when you operate like that, the colorization action is not triggered (which I use the DOM childlist change as trigger for the color generation..).

Now I have one [pull request](https://github.com/FreshRSS/Extensions/pull/186) to fix it.  Let's see.

And for your question: "As a side question: what colors are being used for the origins? Can I configure that somewhere?"

It's quite lazy method I am taking, i.e. just using the 'feed name' to generate random color => 'customized color' is not supported...after I consider the effort to be invested vs. the benefit it can bring (and also lazy me..)

Thanks,

Alkarex:
[...]

PR #185, opened 2023-11-23
Add PHPStan and other quality checks

Alkarex:
Similar to FreshRSS core
Contributes to https://github.com/FreshRSS/Extensions/issues/184
Uses https://github.com/FreshRSS/FreshRSS/pull/5901

TODO:

* [x] Check compatibility with PHP 7.4 - PHP 8.3

PR #186, opened 2023-11-24
Fix for issue #183: xExtension-ColorfulList not working when only viewing unread messages 

shinemoon:
Colorization will not be triggered while in some scenario the loaded page will not have any 'new' DOM structure change, and one 'init' colorization can solve this.


Alkarex:
Remember to also bump the version number in the metadata

shinemoon:
> Remember to also bump the version number in the metadata

Done, and please help to review further.

Alkarex:
Thanks 👍🏻  If you can, please test this new version https://github.com/FreshRSS/Extensions/pull/185 (and tests of other extensions also welcome)

Issue #187, opened 2023-11-25
[Bug] xExtension-ColorfulList: JavaScript error on pages that have no feed articles


math-GH:
ping @shinemoon 

Extension: xExtension-ColorfulList

FreshRSS version: current 1.22.2-dev edge

JavaScript error in the browser console (tested with Firefox 120 on Win10)
![grafik](https://github.com/FreshRSS/Extensions/assets/1645099/c3c8dd25-bf8a-4688-9394-e88c91ec2191)

Error appears on every page, that does not show the feeds.
No error: normal view
Error: subscription manage page. Or: Extension page

shinemoon:
Fix created as above... please help to check. thx.

math-GH:
fine

PR #188, opened 2023-11-25
[Fix] for xExtension-ColorfulList: JavaScript error on pages that have no feed articles #187

shinemoon:
Add pre-check/skip while not in the main page (e.g. configure page and so on) where no valid list will be loaded to avoid annoying error prints....

[[Bug] xExtension-ColorfulList: JavaScript error on pages that have no feed articles](https://github.com/FreshRSS/Extensions/issues/187)


xExtension-ColorfulList/static/script.js
```diff
@@ -1,29 +1,26 @@
-'use strict';
```
math-GH, line 1: I prefer to keep this line

Alkarex:
I merged a bit too fast :-) It lacks at least a metadata version dump

shinemoon:
> I merged a bit too fast :-) It lacks at least a metadata version dump

Thanks Alkarex, but there should be already v0.3..i.e. 1 initial version (0.1) + 2 tickets, so I would say it's expected version?

Issue #190, opened 2023-12-09
Invidious Extension forked and updated


tunbridgep:
Hi, I have made a more up-to-date version of the Invidious extension, forked from the cn-tools version which was in turn forked from Korbeks version.

You can see it here:

https://github.com/tunbridgep/freshrss-invidious

Would this be worthwhile replacing the one in the readme?

Alkarex:
Hello,
Looks good, but when possible, I would like to encourage fixes to be made by pull requests instead of forks, and even if the parent is not responsive, I think it is a good etiquette to send a PR.
Ping @cn-tools for @Korbak for potential merge and/or giving each-other the possibility to contribute to the same repository.

Alkarex:
Readme updated in https://github.com/FreshRSS/Extensions/pull/192

tunbridgep:
I did try getting in touch with them beforehand, I got no response.

Thanks for completing this.

Issue #150, opened 2023-03-05
xExtension-TwitchChannel2RssFeed


babico:
Hi, I made a Twitch plugin using the Youtube plugin of @cn-tools , I wanted to let you know if anyone would like to use it.

**[xExtension-TwitchChannel2RssFeed](https://github.com/babico/xExtension-TwitchChannel2RssFeed)**

PR #192, opened 2023-12-13
added infos about new extensions

cn-tools:
this extensions are not from me but i think they should be referenced.

#150 
#163 
#190


Frenzie (approved):
Fine by me

PR #194, opened 2023-12-28
Update extensions.json: upgrade TogglableMenu to 1.2

nicofrand:


Frenzie:
The repo should normally update that automatically btw. :-)
https://github.com/FreshRSS/Extensions/blob/d741e01f81b78bc6e66558fafbfb8d8221ac3c98/generate.php

PR #196, opened 2023-12-31
Remove obsolete TT-RSS extension

Alkarex:
It was not working, and has not been maintained in a long time.
A follow-up was started on https://github.com/FreshRSS/FreshRSS/pull/1526 as it would require changes in the core for proper support.
But if anybody was using this extension for anything, reach out :-)


marienfressinaud (approved):
If it doesn't work any more, I think it's better to delete it so as not to create false hopes :+1:

math-GH:
Should we find a way to mark this extension as deprecated? In the extension list and if installed?

Alkarex:
> Should we find a way to mark this extension as deprecated? In the extension list and if installed?

That would be good indeed

PR #197, opened 2024-01-01
xExtension-ShareByEmail: add german localization

Green-Kite:
add german localization

### DIFF
--- a/xExtension-ShareByEmail/i18n/de/shareByEmail.php
+++ b/xExtension-ShareByEmail/i18n/de/shareByEmail.php
@@ -1,0 +1,22 @@
+<?php
+
+return [
+	'share' => [
+		'feedback' => [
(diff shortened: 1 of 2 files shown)


math-GH (approved):
didn't test, but the translation looks fine for me

Alkarex:
You might have to bump the version number in
https://github.com/FreshRSS/Extensions/blob/master/xExtension-ShareByEmail/metadata.json

jonastaedcke:
You use „Mail“ and „E-Mail“. I think it would be better to use only one of the two options in all cases.

Frenzie:
Haha, indeed. I'd say go with "Mail," the "E-" part is a bit awkward and I don't think anybody says it anymore.

Alkarex:
Can any of you finalize the PR accordingly?

Frenzie:
I pushed a commit removing the "E-".

Alkarex:
Could you also bump the version number?

PR #207, opened 2024-02-27
Fix access rights for CustomCSS + CustomJS

PsyCh0Taz:
FR : Correction de la gestion des droits.
EN :Rights testing correction

Same test as xExtension-CustomCSS


xExtension-CustomJS/configure.phtml
```diff
@@ -12,7 +12,7 @@
 	</div>
 
 	<div class="form-group form-actions">
-		<?php if (isset($this->permission_problem)) { ?>
+		<?php if (!empty($this->permission_problem)) { ?>
```
Alkarex, line 15:
```suggestion
		<?php if ($this->permission_problem !== '') { ?>
```

And please also in CustomJS at the same time

Alkarex:
Thanks. All merged. Please update your extensions and double-check that all works fine

PR #211, opened 2024-03-03
Update configure.phtml

PsyCh0Taz:
Postponement of the same update as CustomJS


Alkarex:
Moved into https://github.com/FreshRSS/Extensions/pull/211

Issue #206, opened 2024-02-25
[xExtension-ImageProxy] Duplicate images in articles


denis-dysen:
With the proxy enabled, every article that has an image in it, has its first (or only) image duplicated, with the duplicate image being located at the end of the article, and checking the image URLs reveals the duplicate is at the end is the original non-proxied link, while all other images links are proxied.

math-GH:
I investigated it a bit. 

Here are some insights, that roots are in ImageProxy extension:
It wraps the images in a new HTML document "envelope" that completely destroys the HTML: 
```
<div class="text"> 
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body><img>........</body></html>
<figure class="enclosure">
```
(check it with the "show source" in your browser, not with the inspect tool)

In the last new versions of FreshRSS a new feature was implemented that prevent showing duplicated images from the articles as attachment based on the image source. The imageProxy extension exchanges the `src` of the image to the proxy URL, so the mechanisms cannot detect this image as already shown, so the image will be displayed again as attachement with the origin source.

If you manipulated https://github.com/FreshRSS/Extensions/blob/189406e1603b63d60423c0f68798122215ce17cf/xExtension-ImageProxy/extension.php#L129 to `$img->setAttribute('data-proxy-src', $newSrc);` than there are no duplicated images anymore and the `data-proxy-src` is set correctly.

(I do not have any solution for it right now)

Frenzie:
> It wraps the images in a new HTML document "envelope" that completely destroys the HTML:

That's unavoidable. But it's supposed to do something like `$dom->getElementsByTagName('body')->item(0)`.

jonsmy:
Have this problem as well. In the default FreshRSS release feed I get two avatar images: One from my proxy, and another directly linking to github. :/

alithechemist:
Same here: double images, the original and the one from teh proxy.
I'm using this extension just to avoid clients downloading images from third parties websites other that Freshrss, for privacy purposes.

PR #212, opened 2024-03-04
ImageProxy: avoid oversight of wrapping an entire HTML document around image

Frenzie:
In reply to <https://github.com/FreshRSS/Extensions/issues/206#issuecomment-1976649551>.

With special thanks to this comment: <https://www.php.net/manual/en/domdocument.savehtml.php#119767>

### DIFF
--- a/xExtension-ImageProxy/extension.php
+++ b/xExtension-ImageProxy/extension.php
@@ -121,7 +121,7 @@ final class ImageProxyExtension extends Minz_Extension {
 
 		$doc = new DOMDocument();
 		libxml_use_internal_errors(true); // prevent tag soup errors from showing
-		$doc->loadHTML(mb_convert_encoding($content, 'HTML-ENTITIES', 'UTF-8'));
+		$doc->loadHTML(mb_convert_encoding($content, 'HTML-ENTITIES', 'UTF-8'), LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD);
 		$imgs = $doc->getElementsByTagName('img');
 		foreach ($imgs as $img) {
 			if ($img->hasAttribute('src')) {


Frenzie:
Hm, I was just reading more and realized this may not be ideal, though at first glance the output looked okay:

> For an example:
> 
> `<h1>Foo</h1><p>bar</p>`
> 
> will end up as:
> 
> `<h1>Foo<p>bar</p></h1>`
> 
> Easiest workaround is adding root tag yourself and stripping it later:
> 
> `$html->loadHTML('<html>' . $content .'</html>', LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD);`
> 
> `$content = str_replace(array('<html>','</html>') , '' , $html->saveHTML());`

Alkarex:
I have not looked precisely, but a typical way is to `saveHTML($node)`. Try to search our codebase for `->saveHTML(` for some examples

Frenzie:
That's the first thing I tried, as `$dom->getElementsByTagName('body')->item(0)->saveHTML()`.

But I see what you're referring to is actually a for loop appending to a string. I had considered that, but it felt hacky.

Alkarex:
I think it makes a difference in behaviour to pass the node inside the `saveHTML($node)` function

Alkarex:
And no, without loop. See e.g. 

https://github.com/FreshRSS/FreshRSS/blob/da43fff43713daadbf13d05557a2ba14d5166b11/lib/SimplePie/SimplePie/Sanitize.php#L468

Frenzie:
Ah, in that case presumably it should simply be something like:

```php
$output = $dom->getElementsByTagName('body')->item(0);
return $dom->saveHTML($output);
```

I'll test that tomorrow (if I find the time).

PS Here are the other results. I hadn't notice the third was slightly different, whoops.
https://github.com/FreshRSS/FreshRSS/blob/da43fff43713daadbf13d05557a2ba14d5166b11/app/Models/Feed.php#L751-L756
https://github.com/FreshRSS/FreshRSS/blob/da43fff43713daadbf13d05557a2ba14d5166b11/app/Models/Entry.php#L727-L743

Frenzie:
For reference, this is the idea I played with first:
```php
		$body = $doc->getElementsByTagName('body')->item(0);

		$output = '';
		foreach ($body->childNodes as $node) {
				$output .= $doc->saveHTML($node);
		}

		return $output;
```

The sample uses regex to effect the desired result. I will do so as well to limit new string creation.
https://github.com/FreshRSS/FreshRSS/blob/da43fff43713daadbf13d05557a2ba14d5166b11/lib/SimplePie/SimplePie/Sanitize.php#L470-L474

PR #214, opened 2024-03-05
version must be a string that php function version_compare work right

cn-tools:
ref: https://github.com/FreshRSS/FreshRSS/pull/6155


Alkarex:
Updated: https://github.com/FreshRSS/Extensions/blob/b4afee6d0e1d8fb8daa6ac6c324721db68203ba6/extensions.json

PR #215, opened 2024-03-05
ImageProxy: better DOMDocument HTML handling

Frenzie:
With the method in #212 undesired transformations could sometimes be effected.

### DIFF
--- a/xExtension-ImageProxy/extension.php
+++ b/xExtension-ImageProxy/extension.php
@@ -121,7 +121,7 @@ final class ImageProxyExtension extends Minz_Extension {
 
 		$doc = new DOMDocument();
 		libxml_use_internal_errors(true); // prevent tag soup errors from showing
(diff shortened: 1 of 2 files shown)


xExtension-ImageProxy/extension.php
```diff
@@ -136,7 +136,14 @@ public static function swapUris(string $content): string {
 			}
 		}
 
-		return $doc->saveHTML() ?: '';
+		$body = $doc->getElementsByTagName('body')->item(0);
+
+		$output = $doc->saveHTML($body);
+
+		$output = preg_replace('/^<body>/', '', $output);
+		$output = preg_replace('/<\/body>$/', '', $output);
```
Alkarex, line 144:
```suggestion
		$output = preg_replace('/^<body>|<\/body>$/', '', $output);
```

Not tested, but something like that should work and be faster

Frenzie: I would've written it that way, but the example you sent explicitly didn't.

Alkarex:
Do you maybe also want to bump the extension version here?

Issue #219, opened 2024-03-14
Github Action job with an annotation: `The `set-output` command is deprecated and will be disabled soon. 


math-GH:
```
The `set-output` command is deprecated and will be disabled soon.
Please upgrade to using Environment Files.
For more information see:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

![grafik](https://github.com/FreshRSS/Extensions/assets/1645099/6c82c548-b2bb-427d-bc49-cdef3588d280)

Alkarex:
https://github.com/FreshRSS/Extensions/actions/runs/8279809643

https://github.com/FreshRSS/Extensions/blob/dc3fbfb05ea4de32b2e5d1ffd95494c6ea226297/.github/workflows/generate.yml#L26-L30

Alkarex:
Should be fixed by https://github.com/FreshRSS/Extensions/pull/220

PR #220, opened 2024-03-14
Fix deprecated GitHub Actions

Alkarex:
fix https://github.com/FreshRSS/Extensions/issues/219
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/generate.yml
+++ b/.github/workflows/generate.yml
@@ -27,7 +27,7 @@ jobs:
         id: diff
         run: |
           DIFF=$(git diff --numstat -- $FILE | wc -l)
-          echo "::set-output name=DIFF::$DIFF"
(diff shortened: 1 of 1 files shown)


Alkarex:
I am not entirely sure of the syntax, but let's try :-)

Frenzie (approved):
DIFF=$DIFF reads rather redundant to me, but if that's what's necessary.

Alkarex:
Seems to work fine:
https://github.com/FreshRSS/Extensions/actions/runs/8288136129
https://github.com/FreshRSS/Extensions/commit/109246223fcc19143807aab31c82292cddeb3747



> DIFF=$DIFF reads rather redundant to me

The first part is the name of the variable, the second part is the actual value

PR #221, opened 2024-03-17
[TitleWrap] Fix for FreshRSS 1.23.2-dev

math-GH:
https://github.com/FreshRSS/FreshRSS/issues/6207

Improve the CSS


Alkarex:
1.24 is around the corner, but any possibility of making this work with current version 1.23.x as well?

math-GH:
> 1.24 is around the corner, but any possibility of making this work with current version 1.23.x as well?

The change was marginal, so it should work with previous FreshRSS versions too. I tested `1.23.0` and `1.23.1`. Looks fine.

PR #224, opened 2024-03-19
make sure both versions are equal

kevinpapst:
- due to the previous version being an "int" (in metadata.json) we cannot see on screenshots if someone really updated to the latest release or is still using an old version.
- aligned description and title in both files
- remove peer tube reference, as I never used or tested that and never wanted to support it either ... I don't even know what that is or if anyone uses it

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -509,10 +509,10 @@
             "directory": "."
         },
         {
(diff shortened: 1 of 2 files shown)


kevinpapst:
Thanks @Alkarex 👍

PR #237, opened 2024-05-16
xExtension-showFeedID: show IDs of category too

cn-tools:
For an upcoming extension it is good to find out the id of the category easely too


Alkarex:
Some minor lint tests are failing

PR #261, opened 2024-11-29
Update README.md third-party extensions

Rambomst:
Added a third-party extension to the list for custom user stylesheets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,3 +157,7 @@ There are some FreshRSS extensions out there, developed by community members:
 ### By [@LiangWei88](https://github.com/LiangWei88)
 
 * [ArticleSummary](https://github.com/LiangWei88/xExtension-ArticleSummary): A powerful article summarization plugin for FreshRSS that allows users to generate summaries using a language model API conforming to the OpenAI API specification.
+
+### By [@Rambomst](https://github.com/Rambomst)
+
+* [Custom Stylesheet](https://github.com/Rambomst/freshrss-extension-custom-stylesheet): This allows a user to add a custom stylesheet for personal styling tweaks.


Alkarex:
Hello,
Isn't it the same than the built-in core extension "User CSS"?
https://github.com/FreshRSS/FreshRSS/tree/edge/lib/core-extensions/UserCSS

See also https://github.com/FreshRSS/Extensions/pull/256

Rambomst:
> Hello,
> Isn't it the same than the built-in core extension "User CSS"?
> https://github.com/FreshRSS/FreshRSS/tree/edge/lib/core-extensions/UserCSS
> 
> See also https://github.com/FreshRSS/Extensions/pull/256

Similar but not the same.

Rambomst:
The default extension allows the user to define internal CSS via a textarea, it comes a hassle to edit when you get 1000+ lines.

I wanted to be able to have a stylesheet that was easily openable within an IDE that gives context highlighting etc.

The extension I added allows the user to define the location of an external stylesheet.

Frenzie:
The extension isn't set in stone of course. I think it would welcome the ability to provide a URL as well. :-)

> Also the default extension behaves weirdly with CSS @imports, it will import them once, save it, and never run the @import again.

The extension doesn't do anything with `@import`. That's simply regular browser caching based on HTTP headers.

Rambomst:
> The extension isn't set in stone of course. I think it would welcome the ability to provide a URL as well. :-)

If there is interest in this change then I am happy to open a PR to the core extension. Please let me know :)

> > Also the default extension behaves weirdly with CSS @imports, it will import them once, save it, and never run the @import again.
> 
> The extension doesn't do anything with `@import`. That's simply regular browser caching based on HTTP headers.

I performed a hard refresh, I thought that would have busted all the cache related things, no?

Frenzie:
> I performed a hard refresh, I thought that would have busted all the cache related things, no?

My point is that the extension doesn't do any kind of processing. What goes in is what comes out with a `Content-Type: text/css` header, and a `Last-Modified` header indicating whenever you last clicked update. There isn't anything else to it.

Anything under `@import` is dealt with however `@import` is dealt with. My primary suspect in what you say is that this also means it doesn't add any CSP headers.

Rambomst:
I have opened a PR to FreshRSS which introduces this setting to the core UserCSS extension.

https://github.com/FreshRSS/FreshRSS/pull/7107

Alkarex:
Should be added to https://github.com/FreshRSS/Extensions/blob/master/repositories.json as well

PR #264, opened 2024-12-20
ShareByEmail extension: config update

math-GH:
Give information about the configuration of the email system:

After:

(default) PHP mailer is enabled:
grafik

SMTP is enabled:
grafik

Wrong configuration:
grafik

Plus: documentation a bit improved

### DIFF
--- a/xExtension-ShareByEmail/i18n/en/shareByEmail.php
+++ b/xExtension-ShareByEmail/i18n/en/shareByEmail.php
@@ -20,10 +20,10 @@ return [
 		'title' => 'Share an article by email',
 		'manage' => [
 			'mailer' => 'Mailing system',
-			'mail' => 'PHP mailer',
+			'mail' => 'PHP <code>mail()</code>',
 			'smtp' => 'SMTP (send from %s)',
 			'error' => 'Error',
-			'help' => 'Switch PHP mailer/SMTP connection in <kbd>config.php</kbd>: see <a href="https://freshrss.github.io/FreshRSS/en/admins/05_Configuring_email_validation.html#configure-the-smtp-server" target="_blank">documentation</a>'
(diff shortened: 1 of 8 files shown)


xExtension-ShareByEmail/views/shareByEmail/share.phtml
```diff
@@ -3,75 +3,75 @@
...
+				<?= _t('shareByEmail.share.form.to') ?>
+			</label>
 
-            <div class="group-controls">
-                <input
-                    type="email"
-                    name="to"
-                    id="to"
-                    class="long"
-                    placeholder="camille@dupuis-morizeau.com"
```
Alkarex, line 26: What was that? 😮

xExtension-ShareByEmail/i18n/en/shareByEmail.php
```diff
@@ -18,5 +18,12 @@
 		],
 		'intro' => 'You are about to share this article by email: “<strong>%s</strong>”',
 		'title' => 'Share an article by email',
+		'manage' => [
+			'mailer' => 'Mailing system',
+			'mail' => 'PHP mailer',
```
Alkarex, line 23:
```suggestion
			'PHP <code>mail()</code>'
```

Maybe

xExtension-ShareByEmail/views/shareByEmail/share.phtml
math-GH: ping @marienfressinaud

marienfressinaud: Oh that :D The "Dupuis Morizeau" is the fake family that we often use at Framasoft to illustrate our articles or documentation. It changes from the "John Doe" and similar names :)

Alkarex: Ok. It could have been worse... But it is not a good practice to invent potentially valid domain names. Please check https://fr.wikipedia.org/wiki/Example.com

marienfressinaud: Yes, I use `@example.org` or `@example.com` now, but I didn't know that these domains were reserved when I wrote this code :sweat_smile: Good thing that you changed it!

Issue #202, opened 2024-01-15
Update ImageProxy to pass PHPStan Level 9


Alkarex:
https://github.com/FreshRSS/Extensions/pull/185#issuecomment-1826134164 and follow-up of https://github.com/FreshRSS/Extensions/pull/201 , https://github.com/FreshRSS/Extensions/pull/269

ImageProxy does not even pass PHPStan Level 2 at the moment.

Only [a couple of files](https://github.com/FreshRSS/Extensions/blob/15d66caafbfb0dce51f3ce4dfcd2d95d546abfb2/phpstan.dist.neon#L15-L16) are lacking a little update to pass [PHPStan Level 10](https://phpstan.org/user-guide/rule-levels) for all this Extensions repository:

```
./xExtension-ImageProxy/configure.phtml
./xExtension-ImageProxy/extension.php
```

The problem seems to be that ImageProxy saves the configuration parameters in FreshRSS top level instead of one level down dedicated to extensions

How to work:

```sh
git clone https://github.com/FreshRSS/FreshRSS.git
git clone https://github.com/FreshRSS/Extensions.git
cd Extensions/
```

Remove those two lines
https://github.com/FreshRSS/Extensions/blob/15d66caafbfb0dce51f3ce4dfcd2d95d546abfb2/phpstan.dist.neon#L15-L16

```sh
make test-all
```

Alkarex:
Still relevant. See update https://github.com/FreshRSS/Extensions/pull/269

Frenzie:
Some notes on getting PHPStan to work here: https://github.com/FreshRSS/Extensions/pull/209#issuecomment-1971950199

Alkarex:
> Some notes on getting PHPStan to work

See higher up here

PR #270, opened 2025-01-12
Update ImageProxy code

Alkarex:
fix https://github.com/FreshRSS/Extensions/issues/202
Note: breaking changes. Users will have to save their configuration again


Alkarex:
Not tested.

@Frenzie You are welcome to check if you want to add some auto migration of the former way of storing user preferences (storing at user config top level was never meant to be supported).

Frenzie:
The settings are extremely basic so that doesn't seem very necessary but then again there should probably be some kind of warning about it.

Alkarex:
> The settings are extremely basic so that doesn't seem very necessary but then again there should probably be some kind of warning about it.

I am not sure were the warnings should be given, though

Frenzie:
The README seems to have acquired a changelog. Maybe that suffices?
https://github.com/FreshRSS/Extensions/tree/c726958f3fe474c013237c8d885a03225121933c/xExtension-ImageProxy#changelog

There is one other fairly obvious place, but presumably that would require development.
image

Alkarex:
Would you be able to test that it still work as it should?

Frenzie:
The proxy HTTP setting works, but the proxy HTTPS setting doesn't seem to.

Alkarex:
Please try again

Alkarex:
Super. Let's made those changes in another PR. Anyone?

PR #275, opened 2025-01-16
Refactor extension QuickCollapse

Alkarex:
Using the `js_vars` hook instead of the complicated and partially wrong controller+view approach to inject the JavaScript variables.
Looked at due to https://github.com/kalvn/freshrss-mark-previous-as-read/issues/4 which seems to have used it as reference


Alkarex:
Anybody using this extension around there to give it a try?
@romibi @marienfressinaud @mkorthof

romibi:
Added it to my todo list to try it out …
Maybe I'll find time to do it this weekend … but I can't promise for sure …

Alkarex:
Ok 👍🏻 Seems to work, so let me merge so you can more easily test by just updating the extension

romibi:
Seems to work fine.

PR #291, opened 2025-02-08
[xExtension-YouTube] add player autosize feature

intriguedlife:
Hi all,

I implemented this functionality that allows the youtube iframe's size to be automatically determined.

Hopefully others can enjoy this functionality as well.

Best regards

### DIFF
--- a/xExtension-YouTube/extension.php
+++ b/xExtension-YouTube/extension.php
@@ -10,6 +10,10 @@
 final class YouTubeExtension extends Minz_Extension
 {
 	/**
+	 * Whether we set the Youtube iframe to autosize
+	 */
+	private bool $autoSize = false;
+	/**
 	 * Video player width
 	 */
 	private int $width = 560;
@@ -32,6 +36,8 @@ final class YouTubeExtension extends Minz_Extension
 	#[\Override]
 	public function init(): void
 	{
+		Minz_View::appendStyle($this->getFileUrl('style.css', 'css'));
+	
 		$this->registerHook('entry_before_display', [$this, 'embedYouTubeVideo']);
 		$this->registerHook('check_url_before_add', [self::class, 'convertYoutubeFeedUrl']);
 		$this->registerTranslates();
@@ -63,6 +69,11 @@ final class YouTubeExtension extends Minz_Extension
 			return;
 		}
 
+		$autoSize = FreshRSS_Context::userConf()->attributeBool('yt_autosize');
+		if ($autoSize !== null) {
+			$this->autoSize = $autoSize;
+		}
+
 		$width = FreshRSS_Context::userConf()->attributeInt('yt_player_width');
 		if ($width !== null) {
 			$this->width = $width;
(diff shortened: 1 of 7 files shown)


xExtension-YouTube/static/style.css
```diff
@@ -0,0 +1,5 @@
+.yt_autosize_true {
+    height: 100%;
```
Frenzie, line 2: I'd skip the height. ^_^

intriguedlife: It looks like this when i remove the height
image

Frenzie: That would mean there's already a height on it. How about height: auto or initial?

intriguedlife: Yeah you are right. There was height on it.
I made some adjustments. It now works with height removed.

I also discovered that the inline styling attribute seems to get emptied/cleared. Is that intentional?

Frenzie: I believe most such things are stripped from what's in the feed, yes.

intriguedlife: Ok. Good to know. Thanks.

Alkarex:
Any comment @kevinpapst ?

xExtension-YouTube/extension.php
```diff
@@ -180,11 +200,14 @@ public function getHtmlPeerTubeContentForLink(FreshRSS_Entry $entry, string $lin
 	public function getHtml(FreshRSS_Entry $entry, string $url): string
 	{
 		$content = '';
+	
+		if ($this->autoSize) {
```
kevinpapst, line 204: You removed the class `youtube-plugin-video`. Can you re-add it, to not break existing customizations?

xExtension-YouTube/extension.php
```diff
@@ -180,11 +200,14 @@ public function getHtmlPeerTubeContentForLink(FreshRSS_Entry $entry, string $lin
...
+			$iframe_attribute = 'class="yt_player_autosize"';
+		} else {
+			$iframe_attribute = 'width="' . $this->width . '" height="' . $this->height . '"';
```
kevinpapst, line 207: I can't remember why the style attribute was used, but I am almost certain, that I added it on purpose. 
It's gone now.

xExtension-YouTube/i18n/de/ext.php
```diff
@@ -2,6 +2,7 @@
 
 return array(
 	'yt_videos' => array(
+		'autosize' => 'Autogröße des Players',
```
kevinpapst, line 5:
```suggestion
		'autosize' => 'Automatische Größe des Players',
```

kevinpapst (changes requested):
Thanks for sharing this 👍  will test it later on.
Could you please have a look at my comments.

JamBalaya56562:
Thanks for the original implementation and discussion. I have picked this work up in #488, rebased it conceptually onto the current YouTube extension, and incorporated the review feedback here as well as the CSP-related findings from #382.

Since the replacement is now available for review in #488, I think this PR can be closed as superseded once the new PR is accepted.

kevinpapst:
@Alkarex I guess this one can be closed then

PR #320, opened 2025-05-11
Various improvements for ShowFeedID

Inverle:
- Fix loading of script (ensure page has loaded first, before doing anything)
- Add toggling functionality instead of just showing (show/hide IDs)
- Fix HTML injection / potential XSS after showing IDs (due to use of textContent inside innerHTML)
- i18n localization for Polish
- Better styling for button when showing only feeds with errors
- Minor refactor

### DIFF
--- a/xExtension-showFeedID/extension.php
+++ b/xExtension-showFeedID/extension.php
@@ -13,8 +13,8 @@ final class ShowFeedIdExtension extends Minz_Extension {
 	}
 
 	/**
-	 * @param array<mixed> $vars
-	 * @return array<mixed>
+	 * @param array<string,mixed> $vars
+	 * @return array<string,mixed>
 	 */
 	public function jsVars(array $vars): array {
 		$vars['showfeedid_i18n'] = [
--- a/xExtension-showFeedID/extension.php
+++ b/xExtension-showFeedID/extension.php
@@ -5,6 +5,18 @@ declare(strict_types=1);
 final class ShowFeedIdExtension extends Minz_Extension {
 	#[\Override]
 	public function init(): void {
-		Minz_View::appendScript($this->getFileUrl('showfeedid.js'));
+		if (Minz_Request::paramString('c') === 'subscription') {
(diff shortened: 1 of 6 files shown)


xExtension-showFeedID/extension.php
```diff
@@ -5,6 +5,22 @@
...
+		if (Minz_Request::paramString('c') === 'subscription') {
+			$this->registerTranslates();
+			$this->registerHook('js_vars', [$this, 'jsVars']);
+			Minz_View::appendScript($this->getFileUrl('showfeedid.js'));
+		}
+	}
+
+	/**
+	 * @param array<mixed> $vars
+	 * @return array<mixed>
```
Alkarex, line 17: It would be nice to be a bit more precise if possible, like `array<string,string|array<string, mixed>>` or something like that

xExtension-showFeedID/static/showfeedid.js
```diff
@@ -1,38 +1,56 @@
...
+
 	const div = document.querySelector('h1 ~ div');
-	const button = document.createElement('Button');
+	const button = document.createElement('button');
 
 	button.classList.add('btn');
 	button.id = 'showFeedId';
-	button.innerHTML = '<img class="icon" src="../themes/icons/look.svg" /> Show IDs';
+	button.innerHTML = '<img class="icon" src="../themes/icons/look.svg" /> <span>' + i18n.show + '</span>';
+	if (location.href.includes('&error=1')) {
```
Frenzie, line 13: No risk of ?error=1? There's also something like `new URLSearchParams(new URL(location.href)).error` btw.

Inverle: > No risk of `?error=1`? There's also something like `new URLSearchParams(new URL(location.href)).error` btw.

The URL is usually `/i/?c=subscription&error=1`, so in order for that to happen, it would have to be modified manually. But I'll make the change.

PR #321, opened 2025-05-11
Prevent overwrite of js_vars, and try to avoid name conflicts

Inverle:
Stops QuickCollapse from breaking other extensions

### DIFF
--- a/xExtension-QuickCollapse/extension.php
+++ b/xExtension-QuickCollapse/extension.php
@@ -15,13 +15,14 @@ final class QuickCollapseExtension extends Minz_Extension {
 	/**
 	 * @return array<string, string|array<string, string>>
 	 */
-	public function jsVars(): array {
-		return [
+	public function jsVars(array $vars): array {
+		$vars['quick_collapse'] = [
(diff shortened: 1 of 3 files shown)


Frenzie (approved):

xExtension-QuickCollapse/extension.php
```diff
@@ -13,7 +13,8 @@ public function init(): void {
 	}
 
 	/**
-	 * @return array<string, string|array<string, string>>
+	 * @param array<mixed> $vars
+	 * @return array<mixed>
```
Alkarex, line 17: It would be nice to be a bit more precise if possible, like `array<string, string|array<string, mixed>>` or something like that

Inverle: I just copied it from https://github.com/FreshRSS/Extensions/blob/ec349ee5a6c77bcc02a4b7b6cb7926887b8e0809/xExtension-ReadingTime/extension.php#L46-L59

Alkarex: Let's go for at least `array<string,mixed>`, and it should be improved elsewhere as well

PR #331, opened 2025-06-01
ShowFeedID improvements v2

Inverle:
- Only show the button on the correct pages, instead of also incorrectly showing it on the `/i/?c=subscription&a=feed` page
- More consistent styling for the button

### DIFF
--- a/xExtension-showFeedID/static/showfeedid.js
+++ b/xExtension-showFeedID/static/showfeedid.js
@@ -5,16 +5,25 @@ window.addEventListener("load", function () {
 	const i18n = context.extensions.showfeedid_i18n;
 
 	const div = document.querySelector('h1 ~ div');
-	const button = document.createElement('button');
+	const button = document.createElement('a');
 
 	button.classList.add('btn');
+	button.classList.add('btn-icon-text');
 	button.id = 'showFeedId';
 	button.innerHTML = '<img class="icon" src="../themes/icons/look.svg" /> <span>' + i18n.show + '</span>';
-	if (new URLSearchParams(location.search).get('error')) {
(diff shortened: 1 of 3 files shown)


xExtension-showFeedID/static/showfeedid.js
```diff
@@ -5,17 +5,26 @@ window.addEventListener("load", function () {
 	const i18n = context.extensions.showfeedid_i18n;
 
 	const div = document.querySelector('h1 ~ div');
-	const button = document.createElement('button');
+	const button = document.createElement('a');
```
Alkarex, line 8: Any reason for an `<a>` and not for instance a `<span>` if it is not actually a link?

Alkarex: As an example https://github.com/FreshRSS/FreshRSS/pull/7641
When possible, I like the idea of using `<button>` instead of `<a>` when the UI is actually a button.
We use `<button>` several other places

Inverle: Don't forget that there are many other `<a>` that are styled like a button, for example the "Subscription management" button on the main page, the actualize button, the configuration dropdown.....

Inverle: Also, the ability to middle click on the "button" in order to open the link in a new tab is lost after the change, so that may count as a UX degradation, considering that you are able to do that in other places.

(I usually middle click on everything, so it's kind of important for me)

Alkarex: > the ability to middle click on the "button" in order to open the link in a new tab is lost after the change

That is indeed a fair point

PR #332, opened 2025-06-05
Image Camo extension

JadedBlueEyes:
This saves me from running a public proxy to any arbitrary URL, and lets me share infrastructure (with my forgejo instance, for example)

### DIFF
--- a/xExtension-ImageCamo/i18n/fr/ext.php
+++ b/xExtension-ImageCamo/i18n/fr/ext.php
@@ -1,0 +1,20 @@
+<?php
+
+return array(
+	'imagecamo' => array(
+		'proxy_url' => 'URL du proxy Camo',
+		'proxy_url_help' => 'L\'URL de base de votre instance camo (ex: https://camo.example.com)',
+		'hmac_key' => 'Clé HMAC',
+		'hmac_key_help' => 'La clé secrète partagée utilisée pour signer les URLs (doit correspondre à votre configuration camo)',
+		'encoding' => 'Encodage d\'URL',
+		'encoding_help' => 'Choisissez le format d\'encodage d\'URL supporté par votre instance camo',
+		'encoding_base64_desc' => 'recommandé, URLs plus courtes',
+		'encoding_hex_desc' => 'URLs plus longues, insensible à la casse',
+		'scheme_http' => 'Proxifier les images HTTP',
+		'scheme_https' => 'Proxifier les images HTTPS',
+		'scheme_default' => 'Proxifier les URLs relatives au protocole',
+		'scheme_include' => 'Inclure http*:// dans l\'URL',
+		'security_notice_title' => 'Avis de sécurité',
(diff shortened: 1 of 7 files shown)


xExtension-ImageCamo/i18n/fr/ext.php
```diff
@@ -0,0 +1,20 @@
...
+		'hmac_key' => 'Clé HMAC',
+		'hmac_key_help' => 'La clé secrète partagée utilisée pour signer les URLs (doit correspondre à votre configuration camo)',
+		'encoding' => 'Encodage d\'URL',
+		'encoding_help' => 'Choisissez le format d\'encodage d\'URL supporté par votre instance camo',
+		'encoding_base64_desc' => 'recommandé, URLs plus courtes',
+		'encoding_hex_desc' => 'URLs plus longues, insensible à la casse',
+		'scheme_http' => 'Proxifier les images HTTP',
+		'scheme_https' => 'Proxifier les images HTTPS',
+		'scheme_default' => 'Proxifier les URLs relatives au protocole',
+		'scheme_include' => 'Inclure http*:// dans l\'URL',
```
Alkarex, line 16:
```suggestion
		'proxy_url_help' => 'L’URL de base de votre instance camo (ex: https://camo.example.com)',
		'hmac_key' => 'Clé HMAC',
		'hmac_key_help' => 'La clé secrète partagée utilisée pour signer les URLs (doit correspondre à votre configuration camo)',
		'encoding' => 'Encodage d’URL',
		'encoding_help' => 'Choisissez le format d’encodage d’URL supporté par votre instance camo',
		'encoding_base64_desc' => 'recommandé, URLs plus courtes',
		'encoding_hex_desc' => 'URLs plus longues, insensible à la casse',
		'scheme_http' => 'Proxifier les images HTTP',
		'scheme_https' => 'Proxifier les images HTTPS',
		'scheme_default' => 'Proxifier les URLs relatives au protocole',
		'scheme_include' => 'Inclure http*:// dans l’URL',
```

Alkarex:
Ping @Frenzie

JadedBlueEyes:
The software I'm using is https://github.com/cactus/go-camo. 

The main difference compared to the existing image proxies it means you're not just running an open proxy - which isn't really the greatest security idea.

It's a separate extension primarily because of the config, although it could have the image URL replacement logic shared. It's not that complex though, so I didn't think it was worth it.

Frenzie:
> The main difference compared to the existing image proxies it means you're not just running an open proxy - which isn't really the greatest security idea.

Mind, the Apache config in the README contains this:
```
  AuthType Basic
  AuthName "Proxy - Authorized Users ONLY"
  AuthBasicProvider file
  AuthUserFile /etc/apache2/htpasswd/users
  Require valid-user
```
Which is to say, any proxy is only as open as you want.

> It's a separate extension primarily because of the config, although it could have the image URL replacement logic shared. It's not that complex though, so I didn't think it was worth it.

From a maintenance perspective that sounds a bit backwards to me, but perhaps from a UI perspective it could become confusing.

In any event, that particular logic is basically the only logic likely to change anytime soon (unless someone wants to add support for some different format like Camo).

PR #344, opened 2025-07-31
Prevent HTML injection by sanitizing extension values

Inverle:
Currently any third-party extension repository can edit their `metadata.json` to insert something like `<meta http-equiv=refresh content=0;url=//google.com>` to redirect the Extensions page of every FreshRSS instance to `google.com`.

This fix won't require users to update FreshRSS.


Inverle:
> Also possible to inject something like  to collect instance URLs (that would be considered a tracking pixel)

Correction: CSP would prevent this due to `default-src 'self'` and no `img-src` rule

PR #364, opened 2025-09-13
Add extension for unsafe autologin

Inverle:
See: https://github.com/FreshRSS/FreshRSS/pull/7958

### DIFF
--- a/xExtension-UnsafeAutologin/Controllers/authController.php
+++ b/xExtension-UnsafeAutologin/Controllers/authController.php
@@ -1,0 +1,66 @@
+<?php
+declare(strict_types=1);
+
+class FreshExtension_auth_Controller extends FreshRSS_auth_Controller {
+	private static function redirectFormLogin(): void {
+		if (FreshRSS_Auth::hasAccess()) {
+			Minz_Request::forward(['c' => 'index', 'a' => 'index'], true);
+			return;
+		}
+		Minz_Request::forward(['c' => 'auth', 'a' => 'formLogin']);
+	}
+
+	public function loginAction(): void {
+		if (FreshRSS_Context::systemConf()->auth_type !== 'form') {
+			parent::loginAction();
+			return;
+		}
+
+		$username = Minz_Request::paramString('u');
+		$password = Minz_Request::paramString('p', plaintext: true);
+
+		if ($username === '' || $password === '') {
+			self::redirectFormLogin();
+			return;
+		}
+
+		if (!FreshRSS_user_Controller::checkUsername($username) || !FreshRSS_user_Controller::userExists($username)) {
+			Minz_Request::bad(
+				_t('feedback.auth.login.invalid'),
+				['c' => 'index', 'a' => 'index']
+			);
+			return;
+		}
+
+		$config = get_user_configuration($username);
+
+		$s = $config->passwordHash ?? '';
+		$ok = password_verify($password, $s);
+
+		if ($ok) {
+			FreshRSS_Context::initUser($username);
+			FreshRSS_FormAuth::deleteCookie();
+			Minz_Session::regenerateID('FreshRSS');
+			Minz_Session::_params([
+				Minz_User::CURRENT_USER => $username,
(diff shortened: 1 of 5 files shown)


xExtension-UnsafeAutologin/Controllers/authController.php
```diff
@@ -0,0 +1,75 @@
+<?php
+declare(strict_types=1);
+
+class FreshExtension_auth_Controller extends FreshRSS_auth_Controller {
```
Alkarex, line 4: I am worried about the maintenance of this approach based on copying and overriding.
Would you be able to find another approach based on an extension hook (a new one)?
This way, it would both made this part more realistic to maintain and maybe also enable other custom login types

Inverle: > maybe also enable other custom login types

This isn't really a login type, but an extension of form login.
I'm aware of this TODO:

https://github.com/FreshRSS/FreshRSS/blob/128c375fc9de5119b1963d92dc9cd3e423053111/app/Models/Auth.php#L96-L98

Maybe some sort of hook should be put here?

https://github.com/FreshRSS/FreshRSS/blob/128c375fc9de5119b1963d92dc9cd3e423053111/app/Controllers/authController.php#L124-L125

Alkarex: > This isn't really a login type, but an extension of form login

Indeed. But maybe what about making it a new extension-define login type?

Inverle: > Maybe some sort of hook should be put here?

By the way I meant somewhere above not specifically in the POST if condition :p

> Indeed. But maybe what about making it a new extension-defined login type?
> Edit: and which could forward to the Web form login if it does not have the required parameters to log in

That would be a little more complicated than the current implementation.

Alkarex: Maybe be a hook around:
https://github.com/FreshRSS/FreshRSS/blob/128c375fc9de5119b1963d92dc9cd3e423053111/app/Models/Auth.php#L57
either inside or outside the `form` case

Inverle: I looked into it and I don't think a hook is a viable solution, at least not if we want to keep full compatibility.
`accessControl()` gets called only if `giveAccess()` fails, which is not desirable in this case since unsafe autologin could be used for account switching, and that is not possible with this hook if the user is already logged in.

I don't think it makes sense to put a hook in `loginAction()` either, since Minz already has an option for extending the controller.
If the auth controller ever gets removed then it would be enough to just change the extended class from `FreshRSS_auth_Controller` to `FreshRSS_ActionController` and the extension would keep working.

Inverle: Ping @Alkarex

Inverle: > What do you mean with full compatibility?

I meant:

>  unsafe autologin could be used for account switching

with two bookmarks for instance

Inverle: > Could you tell a little bit what you have tried?

Implementing a hook in `accessControl()` as suggested, but then I realized it's not the correct place at all (read above).

And I believe placing a hook directly in `formLoginAction()` wouldn't be easier or harder to maintain than the current approach.

xExtension-UnsafeAutologin/LICENSE
Alkarex: Should be enough with the default license at the root
https://github.com/FreshRSS/Extensions/blob/main/LICENSE

Inverle:
Tests failed on push 🤔
https://github.com/FreshRSS/Extensions/actions/runs/19612890254/job/56161299671

Alkarex:
> Tests failed on push 🤔 https://github.com/FreshRSS/Extensions/actions/runs/19612890254/job/56161299671

Fixed in https://github.com/FreshRSS/Extensions/pull/396
Tests welcome

PR #382, opened 2025-10-27
[xExtension-YouTube] Add referrer policy to iframe in extension.php

QrchackOfficial:
Hi,

I've been seeing Error 153 / Video player configuration error just like reported here:
https://forum.obsidian.md/t/youtube-video-embed-does-not-work-video-player-configuration-error/103181

Added referrer policy as taken from youtube embed iframe via share button on youtube.com, seems to be working nicely.

Additionally, it seems to work just fine with width/height set directly on the iframe element, no need to duplicate with inline style - this is not in the PR, but observed and tested.

### DIFF
--- a/xExtension-YouTube/metadata.json
+++ b/xExtension-YouTube/metadata.json
@@ -2,7 +2,7 @@
 	"name": "YouTube Video Feed",
 	"author": "Kevin Papst, Inverle",
 	"description": "Embed YouTube feeds inside article content.",
-	"version": "1.0.0",
+	"version": "1.1.0",
 	"entrypoint": "YouTube",
 	"type": "user"
 }


Inverle:
> Additionally, it seems to work just fine with width/height set directly on the iframe element, no need to duplicate with inline style

Yes the `style` attribute is also redundant due to CSP, but it was put there for [some reason](https://github.com/FreshRSS/Extensions/pull/291#discussion_r1963960296).
It would be good to remove it since it's spamming the console log with errors:
```
Content-Security-Policy: The page’s settings blocked an inline style (style-src-attr) from being applied because it violates the following directive: “default-src 'self'”
```

abodnar:
I'm still running into this even after updating my extension to the latest code. I've checked the rendered html and it has the `referrerpolicy="strict-origin-when-cross-origin"` in the iframe. 

Is there anything else that could be causing this to happen? or I can provide to help with debugging?

abodnar:
> Look at the Network tab in DevTools (F12) and see in the youtube iframe request if your browser is sending a Referer header.
It's possible YT is doing something else too though (refreshing the iframe a few times may work or not)

Looks like Safari does not sender a referrer header.

Screenshot 2025-10-31 at 10 30 44 AM

I tried this in Chrome and it did work correctly and did send the header.

PR #384, opened 2025-10-30
Add third party extension xExtension-FeedDigest

fengchang:
Hello, I developed a new extension to summarize articles in FreshRSS. It use OpenAI compatible APIs, need user to provide their own key. Please take a look.


Alkarex:
Hello,
Please update https://github.com/FreshRSS/Extensions/blob/master/repositories.json as well

PR #404, opened 2025-12-06
Remove togglable menu extension from list

Inverle:
This is now implemented in FreshRSS instead: https://github.com/FreshRSS/FreshRSS/pull/8201


Inverle:
ping @nicofrand

math-GH:
Let's keep this extension in the list for a while because not every FreshRSS instance will be updated soon. The extension list is global for every version

PR #408, opened 2025-12-24
New extension: Youlag

civilblur:
The Youlag FreshRSS extension allow users to browser FreshRSS with a video focused experience.

### DIFF
--- a/repositories.json
+++ b/repositories.json
@@ -112,4 +112,7 @@
 }, {
 	"url": "https://github.com/veverkap/xExtension-karakeep-button",
 	"type": "git"
+},  {
+	"url": "https://github.com/civilblur/youlag",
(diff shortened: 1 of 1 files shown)


extensions.json
Inverle: This is the auto-generated file by GH actions, you should add your repo to `repositories.json` too.

civilblur: My bad, so `repositories.json` is the only file that should be updated? Should I omit all other changes?

Inverle: The workflow will overwrite `extensions.json` later, so it's fine

Inverle: See https://github.com/FreshRSS/Extensions/commit/235fa01185a453a8f870fe12f6e863ba9ffee5c3

PR #458, opened 2026-04-15
New LLM Classification extension

Alkarex:
Extension to automatically tag incoming articles based on a prompt sent to an LLM.
Require:
* https://github.com/FreshRSS/FreshRSS/pull/8696
* https://github.com/FreshRSS/FreshRSS/pull/8700


Alkarex:
For the background, I have (and have had) a few projects at work where such a feature is needed.
The projects have been using their custom extensions so far, but here is a generic one.

xExtension-LlmClassification/i18n/en/ext.php
```diff
@@ -0,0 +1,48 @@
...
+
+Title: {title}
+Author: {author}
+Date: {date}
+URL: {url}
+Feed: {feed_name} ({feed_url})
+Existing tags: {tags}
+
+Content:
+{content}',
```
Frenzie, line 50: Shouldn't it be split into system and user message?

Alkarex: Yes, there is a system prompt before:
https://github.com/FreshRSS/Extensions/blob/b4abb6d570eba19a4001af73979526088472204c/xExtension-LlmClassification/extension.php#L87-L103

Frenzie: I see. In that case the "classify the following article" in the user prompt seems a bit redundant. But why not editable?

Alkarex: The user prompt is just an example. I expect that the sentence *Classify the following article* will be replaced or expanded by the user. But it works as-is out of the box.
The system-prompt is not editable (due to the need of constraints on the inputs and outputs), but the user-prompt is.

Frenzie: I'm not entirely convinced, but (for now) it's not for me anyway. :-)

Frenzie (approved):

xExtension-LlmClassification/i18n/en/ext.php
Alkarex: I am not sure I understood your suggestion or comment. The user prompt is editable.

Frenzie: I'm mostly referring to the first sentence, the `You are a classification assistant. Analyse the article provided by the user` part in particular seems like it's potentially too generic to get good results.

> You are an expert media analyst. Analyze the article below for tone. (tags would be very positive / positive / neutral / negative / very negative for example)

> You are a professional programmer. Analyze the article for relevant programming concepts/languages.

> You are a professional security researcher. Analyze the article for the degree of urgency.

I assume something along those lines will also work in the user prompt, but then it'll be somewhat in competition with the system prompt, and if nothing else it'll use more tokens for the same result.

Alkarex: Right. We can reduce the system prompt a bit. I will also include a response format constraint

Alkarex: Done

Alkarex:
Related:
* https://github.com/FreshRSS/FreshRSS/discussions/6434
* https://github.com/FreshRSS/FreshRSS/discussions/6649
* https://github.com/FreshRSS/FreshRSS/discussions/7242
* https://github.com/FreshRSS/FreshRSS/discussions/7680
* https://github.com/FreshRSS/FreshRSS/discussions/8788

PR #459, opened 2026-04-15
Fix freshrss-greader-redate repo

Alkarex:
The extension moved address some time ago. The extension is archived though, so it might have to be removed (I have not tested whether it is still working / needed).

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -222,6 +222,17 @@
             "directory": "."
         },
         {
+            "name": "GReader Redate",
+            "author": "Julien Avérous",
+            "description": "Use published date instead of fetching date.",
+            "version": "1.3",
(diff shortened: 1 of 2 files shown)


Alkarex:
Ping @javerous

javerous:
Hi @Alkarex ! Sorry, I forgot to make the change, but I actually thought we removed it from the list: the extension isn't needed anymore (that said, I don't remember why :s). At least I don't use it anymore with Reeder, and I don't have the initial dates issues anymore.

javerous:
It was part of the conversation on this PR: https://github.com/FreshRSS/FreshRSS/pull/7034

Alkarex:
Ok, I will merge it here so we have a trace of the archived location, and then remove it in another PR later (PR welcome)

PR #466, opened 2026-05-25
Add Inoreader extension

caglaryalcin:


Alkarex:
* Feel free to add it to the readme as well
* You may want to check https://framagit.org/nicofrand/xextension-threepanesview and consider synergies
* I would recommend changing name to avoid the trademark and being closer to what the extension does (I originally though it was some kind of API bridge)

PR #470, opened 2026-06-01
[LLM Classification] Skip duplicate LLM calls on no-op feed updates

tcx4c70:
Some RSS sources re-publish existing articles with no semantic change (reformatted author, tweaked date, new enclosure attribute, etc.). FreshRSS detects these as updated entries via hash mismatch and calls EntryBeforeInsert again, which previously triggered another LLM API call for every such pseudo-update.

The extension now stores a SHA-1 of the prompt it sent (plus the exact list of tags it assigned) on each classified entry, under the 'llm_classification' attribute namespace. On a feed update of an already-classified entry:

- if the new prompt hashes to the same value, the prior tags are restored and no LLM call is made;
- otherwise, behaviour follows the new 'Re-classify when content changes' toggle (default on): call the LLM and refresh, or keep the prior tags untouched.

When re-classifying, the prior tag list is used to remove only those exact tags (instead of the previous prefix-based heuristic), so manual tags sharing the prefix are preserved.


tcx4c70:
After the fix, the LLM API requests drop from 4-8k per day to ~800 per day, which matches the number of new articles in my RSS sources.

截屏2026-06-01 22 19 25

Alkarex:
Thanks and sorry for the delay.
I had another change, I wanted to address first, and my first thought was that it looks like quite many changes for a hopefully simple option, but I am taking a look

Alkarex:
Thanks again. I have made some simplifications, not tested much yet though. Could you please check whether that would still work for you?

xExtension-LlmClassification/extension.php
```diff
@@ -406,13 +426,40 @@ public function classifyEntry(FreshRSS_Entry $entry): FreshRSS_Entry {
 
 		$systemPrompt = $this->getSystemPrompt();
 		$userPrompt = $this->buildUserPrompt($entry);
-		if ($userPrompt === '') {
-			return $entry;
+		$classification = null;
+
+		assert($this->getEntrypoint() !== '');	// For PHPStan // TODO: Fix in parent method
```
Alkarex: https://github.com/FreshRSS/FreshRSS/pull/8952

tcx4c70:
Hi @Alkarex I just push a fix.

tcx4c70:
After the fix, the simplification version works now.

图片
(I installed the simplification version at around 06/30 00:00, and installed the fix version at around 07/01 16:00)

PR #476, opened 2026-06-03
Fix captcha widget regression

### LINKED ISSUE #475: [xExtension-Captcha] not works
I have installed the extension xExtension-Captcha and enabled it and added the public and private key for hcaptcha, to test it I logged out myself but the hcaptch is not displayed in the login page which led to not able to login, I do not know anything about code so I asked AI about the issue gave the logs, full extension copy and it told me to edit the line In extension.php, 
```
simply change:
private function captchaWidget(): string {
to:
public function captchaWidget(): string {
```
and it worked without any issue, so my issue is I do not know anything about code so I am asking is this the reason or did I mess up anything in setting up which led to not working. I am currently using docker to run Freshrss.

and here is AI describing the issue, 

When configuring the xExtension-Captcha extension and enabling it to protect the auth_formLogin page, the visual captcha widget (hCaptcha, Turnstile, or reCAPTCHA) fails to render on the login screen.

The JavaScript API script successfully loads in the <head> (e.g., https://js.hcaptcha.com/1/api.js), and the backend validation runs on POST (resulting in a missing-input-response error in the logs), but the actual HTML <div> for the widget is never injected into the formLogin.phtml template.

Steps to Reproduce:

Install and enable xExtension-Captcha as a system extension.

Configure a provider (e.g., hCaptcha with valid site/secret keys).

Set the extension to protect auth_formLogin.

Navigate to the FreshRSS login page.

Notice that the login button is disabled and the JavaScript loads in the network tab, but the captcha widget is completely missing from the UI.

Root Cause Analysis:
The bug is caused by a PHP visibility modifier conflict in extension.php.

The hook for the widget is registered in init():

PHP
$this->registerHook(Minz_HookType::BeforeLoginBtn, [$this, 'captchaWidget']);
However, the callback function itself is defined as private:

PHP
private function captchaWidget(): string {
Because captchaWidget() is private, the FreshRSS core Minz_ExtensionManager is blocked by PHP from executing it when firing the BeforeLoginBtn hook. It fails silently, returning nothing, which skips the HTML injection entirely.

Conversely, the JavaScript injection still works because the ActionExecute hook points to handleProtectedPage(), which is correctly marked as public.

Proposed Fix:
In extension.php (around line 62), simply change the visibility of the captchaWidget method from private to public.

Change this:

PHP
private function captchaWidget(): string {
To this:

PHP
public function captchaWidget(): string {

Inverle:
Closes https://github.com/FreshRSS/Extensions/issues/475

### DIFF
--- a/xExtension-Captcha/extension.php
+++ b/xExtension-Captcha/extension.php
@@ -42,7 +42,7 @@ final class CaptchaExtension extends Minz_Extension {
 			$csp = $this->loadDependencies();
 			if (!empty($csp)) $controller->_csp($csp);
 		} elseif (Minz_Request::is('user', 'create') && !FreshRSS_Auth::hasAccess('admin')) {
-			if (!CaptchaExtension::initCaptcha()) {
+			if (!$this->initCaptcha()) {
 				return false;
 			}
 		}
@@ -58,7 +58,7 @@ final class CaptchaExtension extends Minz_Extension {
 	/**
 	 * @throws FreshRSS_Context_Exception
 	 */
-	private function captchaWidget(): string {
+	public function captchaWidget(): string {
 		$config = self::getConfig();
 		if (!$this->isProtectedPage()) {
 			return '';
--- a/xExtension-Captcha/README.md
+++ b/xExtension-Captcha/README.md
@@ -36,6 +36,8 @@ If you are having trouble with logging in after configuring the extension, you c
 
 ## Changelog
 
+* 1.0.6 [2026-06-04]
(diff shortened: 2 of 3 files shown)


xExtension-Captcha/extension.php
```diff
@@ -42,7 +42,7 @@ public function handleProtectedPage(Minz_ActionController $controller): bool {
 			$csp = $this->loadDependencies();
 			if (!empty($csp)) $controller->_csp($csp);
 		} elseif (Minz_Request::is('user', 'create') && !FreshRSS_Auth::hasAccess('admin')) {
-			if (!CaptchaExtension::initCaptcha()) {
+			if (!$this->initCaptcha()) {
```
Inverle, line 45: Change is not related to the bug, just for consistency to match the if above

PR #478, opened 2026-06-14
Replace old outdated Invidious extension with updated one.

Geekshere:
Hello, the original invidious extension made by korbak hasn't been updated in 4 years and no longer works, I forked it and fixed it, I created this PR to replace the old link with the new one.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -278,12 +278,12 @@
         },
         {
             "name": "Invidious Video Feed",
-            "author": "Korbak (forked from Kevin Papst)",
+            "author": "Geekshere (forked from Korbak)",
             "description": "Embed YouTube feeds inside article content, but with Invidious.",
-            "version": "1.1",
+            "version": "1.0",
             "entrypoint": "Invidious",
             "type": "system",
-            "url": "https://github.com/Korbak/freshrss-invidious",
+            "url": "https://github.com/Geekshere/freshrss-invidious",
             "method": "git",
             "directory": "xExtension-Invidious"
         },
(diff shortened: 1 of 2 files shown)


Alkarex:
Hello,
Ok. I can see that the [referenced fork](https://github.com/tunbridgep/freshrss-invidious) also has not been updated in 3 years 

Ping @Korbak and @tunbridgep if you have any comment

Alkarex:
Please update https://github.com/FreshRSS/Extensions/blob/main/extensions.json as well

Geekshere:
Okay, I updated it and created another PR :)

extensions.json
```diff
@@ -278,12 +278,12 @@
         },
         {
             "name": "Invidious Video Feed",
-            "author": "Korbak (forked from Kevin Papst)",
+            "author": "Geekshere (forked from Korbak)",
             "description": "Embed YouTube feeds inside article content, but with Invidious.",
-            "version": "1.1",
+            "version": "1.0",
```
Alkarex:
```suggestion
            "version": "1.2",
```

tunbridgep:
Thanks for the ping.

Unfortunately I no longer use Invidious since Google basically made third-party Youtube front-ends non-viable, as a result I haven't used this plugin for a while.

I've been finding more and more RSS related errors and other issues (like subscriptions just not updating at all) from multiple invidious hosts, and I think it's related to them being unable to keep up with Google's constant backend changes.

I think that no longer how much it's updated, it will probably never work correctly and will need constant updates - as long as Google is actively attacking third-parties like this, we're always going to have problems.

I'm happy to have my fork merged into the official repo or whatever, but I don't really intend on updating it anymore. Anyone is free to fork it and do whatever they want with it, though.

I could be open to maintaining it more, actually. Maybe I should check out Invidious again since I haven't used it in a couple of years. Perhaps the situation has changed.

PR #481, opened 2026-06-18
Replace old outdated Invidious extension with updated one (repository)

Geekshere:
>Please update https://github.com/FreshRSS/Extensions/blob/main/extensions.json as well

Here is another PR with the updated repository


Alkarex:
Merged in https://github.com/FreshRSS/Extensions/pull/478

PR #487, opened 2026-07-05
Migrate markdownlint-cli to markdownlint-cli2 (single config file)

JamBalaya56562:
## What

Migrate the Markdown linter from `markdownlint-cli` to **`markdownlint-cli2`**, and consolidate its configuration into a **single file**.

Previously the setup was spread across three places:

- `.markdownlint.json` — rules
- `.markdownlintignore` — excluded paths
- the `'**/*.md'` glob argument in the npm scripts

These are now merged into one **`.markdownlint-cli2.jsonc`** (`config` + `globs` + `ignores`), and the two old files are removed.

## Why

For consistency with the same migration in the main repository, FreshRSS/FreshRSS#8987. As requested by @Alkarex there:

> Fine with me, at a first glance, but then I would like consistency with the other repo https://github.com/FreshRSS/Extensions so it would be nice to do the same there

## Details

- `package.json`: `markdownlint-cli@^0.49.0` → `markdownlint-cli2@^0.23.0`; the `markdownlint` / `markdownlint_fix` scripts now call `markdownlint-cli2` / `markdownlint-cli2 --fix`.
- **Script names are unchanged**, so the `test` / `fix` compound scripts and the `npm run --silent markdownlint` step in `.github/workflows/tests.yml` need no change.
- `markdownlint-cli2` v0.23.0 bundles the **same markdownlint v0.41.0** already used by `markdownlint-cli` v0.49.0, so linting behaviour is identical — the suite still finds 16 files and reports 0 errors.
- No docs referenced `.markdownlint.json`, so no documentation changes are needed.

## Checklist

- [x] `npm run markdownlint` passes (0 errors, 16 files; `node_modules` and vendored dirs ignored)
- [x] `npm run markdownlint_fix` works (`--fix`), no Markdown files changed
- [x] No behaviour change (same bundled markdownlint 0.41.0)


JamBalaya56562:
@Alkarex applied your review changes from FreshRSS/FreshRSS#8987 here as well:

- Reduced the obvious comments in `.markdownlint-cli2.jsonc`, keeping only the non-obvious note on `globs`.
- Added JSONC to `.editorconfig` (`[*.json]` -> `[*.{json,jsonc}]`).

`npm run markdownlint` still passes (0 errors, 16 files). Both repos stay consistent now. Thanks!

PR #488, opened 2026-07-11
[xExtension-YouTube] Add optional responsive player sizing

JamBalaya56562:
## Summary

- add an opt-in responsive sizing setting for the YouTube and PeerTube players
- preserve the existing `youtube-plugin-video` class and fixed-size behavior by default
- move player sizing out of the inline `style` attribute so it works with FreshRSS CSP
- keep the `referrerpolicy` introduced in #382

This takes over the work from #291 and applies its review feedback to the current implementation.

## Behavior

When automatic sizing is enabled, the player uses the existing `youtube-plugin-video` class together with `yt-player-autosize`, which applies a full-width 16:9 aspect ratio. When it is disabled or unset, the configured `width` and `height` attributes are preserved.

## Validation

- `npm run eslint`
- `npx stylelint "**/*.css"`
- `npm run markdownlint`
- reviewed the generated iframe branches for YouTube, youtube-nocookie, PeerTube, and original-content handling

PHPStan and PHPCS were not run locally because PHP and `make` are unavailable on PATH and Docker Desktop could not start. They are expected to run in GitHub Actions.

### DIFF
--- a/xExtension-YouTube/configure.phtml
+++ b/xExtension-YouTube/configure.phtml
@@ -6,6 +6,13 @@ declare(strict_types=1);
 	<input type="hidden" name="_csrf" value="<?= FreshRSS_Auth::csrfToken() ?>" />
 	<div class="form-group">
 
+		<div class="group-controls">
+			<label class="checkbox" for="yt_autosize">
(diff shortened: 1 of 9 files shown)


kevinpapst:
Thanks @JamBalaya56562 👍

PR #489, opened 2026-07-11
[Share By Email] Preserve special characters in email subjects

### LINKED ISSUE #342: [Share By Email] HTML entities in email's subject
Hello,

I just sent an email to myself with the subject containing a word in "double quotes".
Thunderbird received the message with `&quot;word&quot;`

JamBalaya56562:
## Summary

- preserve special characters in Share By Email subjects instead of XML-encoding them
- bump the Share By Email extension version to 0.3.4

## Root cause

`Minz_Request::paramString()` XML-encodes special characters by default. As a result, a subject such as `"word"` was passed to PHPMailer as `&quot;word&quot;` and displayed that way by email clients.

Reading the subject with `plaintext: true` keeps the original text for PHPMailer, including quotation marks, ampersands, and non-ASCII characters.

Fixes #342

## Validation

- confirmed the default and `plaintext: true` behavior in `Minz_Request::paramString()`
- reviewed the subject data flow from the POST parameter to `ShareByEmail\mailers\Share` and PHPMailer
- confirmed the diff is limited to the Share By Email controller and metadata

PHPStan and PHPCS were not run locally because PHP is unavailable in the local environment. They are expected to run in GitHub Actions.

### DIFF
--- a/xExtension-ShareByEmail/Controllers/shareByEmailController.php
+++ b/xExtension-ShareByEmail/Controllers/shareByEmailController.php
@@ -70,7 +70,7 @@ final class FreshExtension_shareByEmail_Controller extends Minz_ActionController
 
 		if (Minz_Request::isPost()) {
 			$this->view->to = $to = Minz_Request::paramString('to');
-			$this->view->subject = $subject = Minz_Request::paramString('subject');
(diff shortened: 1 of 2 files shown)


Alkarex (approved):

mo74m3ed (commented):
mo74m3ed@hotmail.com

PR #272, opened 2025-01-14
ImageProxy: update text for "Update unspecified scheme"

Frenzie:
See <https://github.com/FreshRSS/Extensions/pull/270#issuecomment-2588343167>.

PR #493, opened 2026-07-12
[Image Proxy] Replace deprecated HTML entity conversion

### LINKED ISSUE #225: [ImageProxy] PHP Deprecated:  mb_convert_encoding()
`PHP Deprecated:  mb_convert_encoding(): Handling HTML entities via mbstring is deprecated; use htmlspecialchars, htmlentities, or mb_encode_numericentity/mb_decode_numericentity instead in /..../extensions/xExtension-ImageProxy/extension.php on line 124`

Image Proxy 0.7.1
FreshRSS 1.23.2-dev
PHP 8.2

JamBalaya56562:
## Summary

- replace the deprecated `mb_convert_encoding(..., 'HTML-ENTITIES', ...)` call with `mb_encode_numericentity()`
- preserve PHP 8.1+ support and the existing `DOMDocument` processing path
- add a dependency-free ImageProxy regression test runner with 19 assertions
- run the regression suite from Composer and GitHub Actions
- bump Image Proxy to 1.0.1

Fixes #225

## Root cause

ImageProxy converts non-ASCII UTF-8 characters to HTML entities before passing article fragments to `DOMDocument::loadHTML()`. Using `HTML-ENTITIES` as an `mb_convert_encoding()` target has been deprecated since PHP 8.2, so every processed article could emit a deprecation warning.

`mb_encode_numericentity()` performs the required non-ASCII-to-entity conversion without using the deprecated pseudo-encoding. This keeps HTML tags and existing entities intact and avoids raising the minimum supported PHP version.

## Regression coverage

The new standalone test runner converts `E_DEPRECATED` notices into exceptions and verifies:

- ASCII, accented Latin text, Japanese, Chinese, emoji, and supplementary-plane characters
- existing named, decimal, and hexadecimal entities without double encoding
- invalid UTF-8 without discarding the complete article
- empty input, HTML documents, fragments, tag soup, and element attributes
- multiple images, `src`, `srcset`, and original-URL data attributes
- HTTP, HTTPS, protocol-relative, encoded, and unsupported URI schemes

## Validation

- `composer run test-image-proxy` via Docker: 19 assertions passed
- PHP syntax checks for the extension and test runner
- `npm run eslint`
- `npx stylelint "**/*.css"`
- `npm run markdownlint`

PHPCS and PHPStan are validated by GitHub Actions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,6 +49,9 @@ jobs:
     - name: PHP_CodeSniffer
       run: composer run-script phpcs
 
+    - name: ImageProxy regression tests
+      run: composer run-script test-image-proxy
+
     - name: Git checkout FreshRSS source code
       uses: actions/checkout@v7
       with:
--- a/composer.json
+++ b/composer.json
@@ -69,9 +69,11 @@
 		"phpcbf": "phpcbf . -p -s",
 		"phpstan": "phpstan analyse .",
 		"phpstan-third-party": "phpstan analyse -c phpstan-third-party.neon .",
+		"test-image-proxy": "php tests/ImageProxyTest.php",
 		"test": [
 			"@php-lint",
 			"@phtml-lint",
+			"@test-image-proxy",
 			"@phpcs",
 			"@phpstan"
 		],
--- a/tests/ImageProxyTest.php
+++ b/tests/ImageProxyTest.php
@@ -1,0 +1,248 @@
+<?php
+
+declare(strict_types=1);
+
+class ImageProxyTestMinzExtension {
+	public function init(): void {
+	}
+
+	public function handleConfigureAction(): void {
+	}
+}
+
+class ImageProxyTestContextException extends RuntimeException {
+}
+
+final class ImageProxyTestConfiguration {
+	/** @param array<string, bool|string> $values */
+	public function __construct(private array $values) {
+	}
+
+	public function attributeBool(string $key): ?bool {
+		$value = $this->values[$key] ?? null;
+		return is_bool($value) ? $value : null;
+	}
+
+	public function attributeString(string $key): ?string {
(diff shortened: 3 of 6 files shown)


Frenzie:
I tried that already and it didn't work very well. I'd much rather just switch to the PHP 8.4 HTML parser.

JamBalaya56562:
Thanks for the context. I found the earlier discussion in #272, but not the exact patch or input that failed.

I tested the `²³⁸Umbrella policy` example mentioned there with both conversions. In isolation, both survived the DOM round trip, so the failure may depend on the surrounding feed HTML.

Could you clarify two points?

1. Do you still have a concrete feed entry or HTML fixture that demonstrates the failure? I would like to preserve it as a regression test.
2. Would you prefer making ImageProxy require PHP 8.4 and replacing `DOMDocument::loadHTML()` entirely with `Dom\HTMLDocument::createFromString()`, or should we retain a fallback for PHP 8.1–8.3?

I agree that switching to the standards-compliant PHP 8.4 parser is preferable to keeping the legacy parser and only suppressing the deprecation warning. Once the compatibility expectation is clear, I can revise this PR accordingly.

Frenzie:
I prefer to do nothing at all until such time as FreshRSS itself raises its version but in lieu of that switching over to PHP 8.4's HTML parser prematurely is still preferable over other changes imho.

JamBalaya56562:
Thanks, understood. I see the rationale for waiting and eventually replacing the parser itself instead of making an interim conversion change.

Even if the conversion change in this PR is not adopted, the regression tests added here might still be useful as a starting point for validating the future PHP 8.4 HTML parser migration. They could potentially be kept or reused separately when that work happens.

Frenzie:
I did some testing and I think this should be fine.

PR #495, opened 2026-07-15
[xExtension-Webhook] Add freshrss username placeholder

HiperDoo:
## Summary

- Add a placeholder to send the FreshRSS username to the webhook

## Root Cause

If a single service consumes all FreshRSS webhooks, and there is more than one registered user, there is no way to distinguish which user the webhook event belongs to.

My scenario: I use n8n to connect to the FreshRSS webhook, it will generate an NTFY notification and it will be sent to the channel according to the username (each user will see their own notifications).

## Images

image
image


Alkarex (approved):

Alkarex:
Thanks!

PR #497, opened 2026-07-24
[LLM Classification] Fix: Strip prefix from existing tags before re-adding in LLM classification

### LINKED ISSUE #496: [LLM Classification] Tag prefix duplication when filtering existing tags
When using LLM Classification with a tag prefix, existing tags that start with the prefix are being re-added with the prefix already applied. This causes the prefix to be duplicated when applyClassification re-applies it.

For example, if the prefix is llm: and an existing tag is llm:technology, it gets re-added as llm:llm:technology instead of just technology.

The issue is in the classifyEntry method where tags are filtered by prefix but the prefix is not stripped before re-adding them to the classification array.

Image

a-p-a-n:
When filtering tags by prefix, the prefix was being preserved on existing tags before they were re-added. This caused the prefix to be duplicated when applyClassification re-applied it. Now we strip the prefix from existing tags before re-adding them, preventing the duplication issue.

Closes #496

### DIFF
--- a/xExtension-LlmClassification/extension.php
+++ b/xExtension-LlmClassification/extension.php
@@ -443,6 +443,9 @@ final class LlmClassificationExtension extends Minz_Extension {
 						$existingTags,
 						static fn(string $tag) => str_starts_with($tag, $prefix)
 					));
+					// Strip the prefix from tags before re-adding them (will be re-applied in applyClassification)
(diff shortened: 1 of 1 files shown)


Alkarex (approved):

Alkarex:
Thanks!

PR #527, opened 2026-09-17
Adds the Heatmap Extension to the community list.

iovictor:
Adds my Heatmap Extension to the community list.

Heatmap: Automatically calculates and displays a temperature for every article in your feeds, helping you cut through the noise by highlighting articles that are highly upvoted or frequently shared across the web.

Cheers!


extensions.json
Inverle: The correct file to edit is `repositories.json`.

